PR #39, opened 2026-03-24
feat(cli): deployment history and version-scoped diff/review

Ruari-Phipps:
## Summary
Adds **`poly deployments`**, extends **`diff`** / **`review`** with hash and **`--before` / `--after` / `--files`**, updates.
Updates **`review`** to be **`review create`** to be similar with **`branch`** commands

## Motivation
Improves visibility into deployed versions and makes comparing local vs remote or named versions consistent in the CLI.

## Changes
- **`poly deployments`** with `--env`, pagination, `--hash`, `--oneline`, `--json`; Rich output with sandbox / pre-release / live badges.
- **`get_deployments`** (API + project): `client_env`, list return shape, tuple with active hashes on **`AgentStudioProject`**.
- **`poly diff` / `poly review`**: optional hash, `--files`, `--before` / `--after`; **`--delete`** on review; shared diff computation.
- **`poly review`** logic moved to `poly review create`
- **`poly revert`** / **`poly format`**: CLI shape updates to be consistent (`--all` removed; format uses **`--files`**).

## Test strategy

## Checklist
- [x] `pytest` passes

## Screenshots / Logs
Screenshot 2026-03-27 at 14 23 31

`--details`
Screenshot 2026-03-27 at 15 35 14

`--env`
Screenshot 2026-03-27 at 15 35 27


src/poly/cli.py
```diff
@@ -2452,6 +2611,60 @@ def docs(
...
+            return
+
+        if version_hash:
+            version_hash = version_hash[:9]
+            version_idx = next(
+                (i for i, v in enumerate(versions) if v.get("version_hash")[:9] == version_hash),
+                None,
+            )
+            if version_idx is None:
+                error(f"Version hash '{hash}' not found.")
```
alanacevedo: should be version_hash

src/poly/cli.py
```diff
@@ -2452,6 +2611,60 @@ def docs(
...
+        versions, active_deployment_hashes = project.get_deployments(client_env=environment)
+
+        if not versions:
+            error("No versions found.")
+            return
+
+        if version_hash:
+            version_hash = version_hash[:9]
+            version_idx = next(
+                (i for i, v in enumerate(versions) if v.get("version_hash")[:9] == version_hash),
```
alanacevedo: could crash is version_hash is None? or this guaranteed to have a value?

maybe

(v.get("version_hash") or "")[:9]

src/poly/cli.py
alanacevedo, line 957: If no subcommand is passed, maybe we should return some warning

src/poly/cli.py
```diff
@@ -1324,103 +1434,186 @@ def status(cls, base_path: str, output_json: bool = False) -> None:
...
+                    for i, v in enumerate(versions)
+                    if (v.get("version_hash") or "")[:9] == after[:9]
+                ),
+                None,
+            )
+            if version_idx is None:
+                error(f"Version hash '{after}' not found.")
+                return
+            if version_idx == len(versions) - 1:
+                error("No previous version found.")
```
alanacevedo, line 1503: `_compute_diff` returns None on error (after printing via `error()`) but I think it could return `{}` from the underlying project methods when there are no changes. The callers at lines ~1546 and ~1633 rely on `diffs is not None` to distinguish the two, which is subtle and easy to get wrong since both are falsy. 

What if we raised an exception and let the callers invoke `error`?

src/poly/cli.py
Ruari-Phipps: Screenshot 2026-04-17 at 14 50 39
Alsready does

alanacevedo (approved):

PR #48, opened 2026-03-26
feat: add anonymous usage telemetry to ADK CLI

christopherosborne-poly:
## What

Adds a lightweight telemetry module so the product team can see exactly who is using the ADK and how.

## How it works

- **New module:** `src/poly/telemetry.py`
  - Single `track(command, ...)` call that fires in a **daemon thread** with a 2-second timeout
  - Never blocks or errors the CLI — all exceptions are silently swallowed
  - Auto-reads `account_id`, `project_id`, `region` from the loaded project context

- **Single integration point:** one block in `_run_command()` in `cli.py` — no logic scattered across individual command handlers

- **Ingest endpoint:** `https://analytics.us-1.platform.polyai.app/ingest/v1/adk`

## What's collected

| Field | Value |
|---|---|
| `command` | e.g. `push`, `pull`, `branch` |
| `subcommand` | e.g. `list`, `switch` (branch actions) |
| `user_email` | from `git config user.email` |
| `user_name` | from `git config user.name` |
| `os_username` | OS login name |
| `hostname` | machine hostname |
| `adk_version` | package version |
| `python_version` | e.g. `3.14.0` |
| `os` | `darwin` / `linux` / `windows` |
| `account_id` | from project.yaml (when available) |
| `project_id` | from project.yaml (when available) |
| `region` | from project.yaml (when available) |
| `dry_run` | `true` only when flag is set |

## Opt-out

Set `POLY_TELEMETRY_DISABLED=1` to disable entirely. Useful for CI/CD pipelines.

## Tests

All 361 existing tests pass.

### DIFF
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -42,6 +42,7 @@ from poly.handlers.interface import (
     REGIONS,
     AgentStudioInterface,
 )
+from poly import telemetry
 from poly.resources import resource_utils
 from poly.project import (
     PROJECT_CONFIG_FILE,
@@ -575,6 +576,28 @@ class AgentStudioCLI:
             logging.basicConfig(level=logging.DEBUG)
         else:
             logging.basicConfig(level=logging.WARNING)
+
+        # --- Usage telemetry (anonymous, fire-and-forget) ---
+        _subcommand = getattr(args, "action", None)
+        _project_ctx: dict = {}
+        if args.command not in {"init", "docs", "completion"}:
+            try:
+                _p = cls.read_project_config(getattr(args, "path", "."))
+                if _p:
+                    _project_ctx = {
+                        "account_id": getattr(_p, "account_id", None),
+                        "project_id": getattr(_p, "project_id", None),
+                        "region": getattr(_p, "region", None),
+                    }
+            except Exception:
+                pass
+        telemetry.track(
+            args.command,
+            subcommand=_subcommand,
+            dry_run=bool(getattr(args, "dry_run", False)),
+            **_project_ctx,
+        )
+        # ---
 
         if args.command == "init":
             cls.init_project(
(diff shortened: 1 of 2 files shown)


HelenGreul (changes requested):
@christopherosborne-poly thanks for putting this together - having visibility into how the ADK is used is a great idea, and I appreciate the clean integration point in `_run_command()`. A few things I'd like us to address before merging:

## Privacy: collected data is not anonymous

The PR title and description describe this as "anonymous usage telemetry," but the code collects `user_email`, `user_name`, `os_username`, and `hostname`. That's personally identifiable information that ties every CLI invocation to a specific person and machine. We should either:

1. **Strip the PII fields entirely** and stick to aggregate/anonymous signals (command, OS, ADK version, account/project IDs), or
2. **If individual-level tracking is a hard product requirement**, rename/rebrand this away from "anonymous," add a clear first-run consent prompt, and link to a privacy notice explaining what's collected and why.

I'd strongly prefer option 1 unless there's a compelling reason we need individual identity.

## No tests

`telemetry.py` is 142 lines of new code with zero test coverage is causing the coverage regression (70.4% → 70.0%). At minimum we should have tests for:

- `_is_disabled()` respects the env var
- `track()` is a no-op when `POLY_TELEMETRY_DISABLED=1`
- `track()` doesn't raise on network failure (the core contract of "never breaks the CLI")
- The payload shape is correct

These are all straightforward to test with `unittest.mock.patch` on `_post` and `subprocess.run`.

## Silent exception swallowing

Every failure path is `except Exception: pass`. I understand the intent - telemetry should never break the CLI - but silently swallowing *all* exceptions makes it very hard to debug issues in production. Could we add `logging.debug(...)` in the except blocks so failures are visible when someone runs with `--verbose`?

## CI failure

The `coverage` check failure is unrelated to your code - it's a 403 from `actions/github-script` trying to post a comment (`Resource not accessible by integration`). That's a workflow permissions issue we should fix separately, but wanted to flag it so you know it's not your fault.

---

Happy to discuss any of these - the core idea is solid, just want to make sure we get the details right before shipping. Thanks!

PR #65, opened 2026-04-07
fix: Add directory walking to account for topics with '/' in name

milesapnash:
## Summary

Fix topic discovery to find YAML files in nested subdirectories by replacing `os.listdir` with `os.walk` in `Topic.discover_resources`.

## Motivation

Topics whose names contain `/` (e.g. `billing/payments`) are stored in subdirectories under `topics/`. The previous implementation used `os.listdir`, which only scanned the top-level directory and silently missed these nested topic files.

## Changes

- `Topic.discover_resources` now recursively walks the `topics/` directory tree to find all `.yaml`/`.yml` files
- Added unit tests for `discover_resources`: basic discovery, nonexistent path, and nested subdirectory cases

## Test strategy

## Checklist

- [ ] `pytest` passes
- [ ] No breaking changes to the `poly` CLI interface (or migration path documented)

### DIFF
--- a/src/poly/resources/topic.py
+++ b/src/poly/resources/topic.py
@@ -160,9 +160,10 @@ class Topic(YamlResource):
         if not os.path.exists(topics_path):
             return discovered_topics
 
-        for file_name in os.listdir(topics_path):
-            if file_name.endswith(".yaml") or file_name.endswith(".yml"):
-                file_path = os.path.join(topics_path, file_name)
-                discovered_topics.append(file_path)
+        for dirpath, _, filenames in os.walk(topics_path):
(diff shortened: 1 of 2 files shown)


Ruari-Phipps (approved):
LGTM! Think we'll have to come up with a better long term solution for this, but for now this does the job

Ruari-Phipps (changes requested):
If it reads a file with this file structure though, does it read the name correctly? Or will it just be after the /

milesapnash:
Will be resolved via topic names, implemented in https://github.com/polyai/adk/pull/94

PR #66, opened 2026-04-08
feat: store project name on init and clean up on API errors

oeisenberg:
## Summary

- Stores the human-readable project name in `project.yaml` on init, 
- Cleans up partially created directories when the API returns an error (e.g. 403 Forbidden, 404 not found).

## Motivation
- Project name was discarded during init — now persisted for downstream use.
- When initialising with an invalid or inaccessible project/account ID, empty directories were left behind.

## Changes

- Added `project_name` field to `AgentStudioProject` dataclass, persisted in `project.yaml` and status file
- On interactive init, the selected project name is captured; on non-interactive init (`--project_id`), the name is resolved via `get_projects()` reverse lookup
- Wrapped `init_project` call in error handling that cleans up `account_id/project_id` directories on `HTTPError` or `SourcererAPIError`
- Friendly error messages for `FORBIDDEN` and `DEPLOYMENT_NOT_FOUND` error codes

## Test strategy

```yaml
project_id: test-bot
account_id: oliver-eisenberg-ws
region: us-1
project_name: New Project
```

```bash
❯ poly init --account_id oliver-eisenberg-ws --project_id non_existant --region us-1 --json
{
  "success": false,
  "error": "Project 'non_existant' not found in account 'oliver-eisenberg-ws'."
}
❯ ls oliver-eisenberg-ws
las-dev-usp
```

```bash
❯ poly init --account_id fake-ws --project_id non_existant --region us-1
Initialising project...
Initializing project fake-ws/non_existant...
Error: Forbidden: you do not have permission to access project 'non_existant' in account 'fake-ws'.
```

## Checklist

- [x] `pytest` passes


src/poly/cli.py
```diff
@@ -1018,37 +1028,69 @@ def init_project(
...
+
+            if error_code == "FORBIDDEN":
+                init_error = (
+                    f"Forbidden: you do not have permission to access "
+                    f"project '{project_id}' in account '{account_id}'."
+                )
+            elif error_code == "DEPLOYMENT_NOT_FOUND":
+                init_error = f"Project '{project_id}' not found in account '{account_id}'."
+            else:
+                init_error = f"API error: {e}"
```
Ruari-Phipps: I don't think any of this should be in CLI

src/poly/project.py
```diff
@@ -343,52 +356,90 @@ def init_project(
...
+
+        except (requests.HTTPError, SourcererAPIError) as e:
+            # Clean up any partially created directories
+            if os.path.exists(project_path):
+                shutil.rmtree(project_path)
+            if os.path.exists(account_path) and not os.listdir(account_path):
+                shutil.rmtree(account_path)
+
+            # Extract error_code from the response body
+            error_code = cls._extract_error_code(e)
```
Ruari-Phipps: If they don't have permission, it shouldn't have written anything?

src/poly/project.py
```diff
@@ -343,52 +356,90 @@ def init_project(
...
+                raise ValueError(
+                    f"Forbidden: you do not have permission to access "
+                    f"project '{project_id}' in account '{account_id}'."
+                ) from e
+            elif error_code == "DEPLOYMENT_NOT_FOUND":
+                raise ValueError(
+                    f"Project '{project_id}' not found in account '{account_id}'."
+                ) from e
+            else:
+                raise ValueError(f"API error: {e}") from e
```
Ruari-Phipps: Maybe something like this can go in the API handler?

src/poly/project.py
oeisenberg: The api_handler property calls save_config() before pull_resources() is invoked:                                  
  
```python                                                                                                         
  @property                                                                                                         
  def api_handler(self):                                                                                            
      if self._api_handler is None:                                                                                 
          self._api_handler = AgentStudioInterface(...)                                                             
      self.branch_id = self._api_handler.branch_id                                                                  
      if self.branch_id:                                                                                          
          self.save_config()  # writes _gen/.agent_studio_config ← happens here                                     
      return self._api_handler         
```                                                                             
So by the time pull_resources() raises FORBIDDEN/DEPLOYMENT_NOT_FOUND, project_path/_gen/.agent_studio_config is  already on disk.

src/poly/cli.py
```diff
@@ -1033,30 +1042,33 @@ def init_project(
...
+                    format=format,
+                    projection_json=projection_json,
+                    on_save=on_save,
+                )
+        except ValueError as e:
+            if output_json:
+                json_print({"success": False, "error": str(e)})
+            else:
+                error(str(e))
+            sys.exit(1)
```
Ruari-Phipps: This is already handled by by in _run_command

src/poly/project.py
```diff
@@ -343,52 +357,64 @@ def init_project(
...
-        utils.save_imports(base_path)
+        except (requests.HTTPError, SourcererAPIError) as e:
+            # Clean up any partially created directories
+            if os.path.exists(project_path):
+                shutil.rmtree(project_path)
+            if os.path.exists(account_path) and not os.listdir(account_path):
+                shutil.rmtree(account_path)
 
-        return project, projection
+            PlatformAPIHandler.translate_http_error(e, project_id, account_id)
```
Ruari-Phipps: Shouldn't this run in the handler itself?

src/poly/project.py
```diff
@@ -343,52 +357,64 @@ def init_project(
...
+            return project, projection
 
-        utils.export_decorators(DECORATORS, base_path)
-        utils.save_imports(base_path)
+        except (requests.HTTPError, SourcererAPIError) as e:
+            # Clean up any partially created directories
+            if os.path.exists(project_path):
+                shutil.rmtree(project_path)
+            if os.path.exists(account_path) and not os.listdir(account_path):
+                shutil.rmtree(account_path)
```
Ruari-Phipps: Have this only around the pull?

src/poly/handlers/interface.py
```diff
@@ -142,7 +144,10 @@ def pull_resources(
             return SyncClientHandler.load_resources_from_projection(
                 projection_json
             ), projection_json
-        return self.sync_client.pull_resources()
+        try:
+            return self.sync_client.pull_resources()
+        except (requests.HTTPError, SourcererAPIError) as e:
+            PlatformAPIHandler.translate_http_error(e, self.project_id, self.account_id)
```
Ruari-Phipps: Maybe have this function just be in the interface if we're using it for both?
Can we use it to wrap all our requests?

oeisenberg: https://github.com/polyai/adk/pull/106

src/poly/handlers/interface.py
```diff
@@ -28,6 +30,32 @@ class AgentStudioInterface:
...
+        Extracts the error_code from the response body and raises a ValueError
+        with a descriptive message. Always raises.
+
+        Args:
+            e: The HTTPError or SourcererAPIError to translate
+
+        Raises:
+            ValueError: Always raised with a user-facing message
+        """
+        error_code = PlatformAPIHandler.extract_error_code(e)
```
Ruari-Phipps: Again should move it in here if its only used here?

Ruari-Phipps (approved):

PR #68, opened 2026-04-10
fix: Raise proper error message for invalid format functions

Ruari-Phipps:
## Summary

Fixes two related issues: untyped or unsupported function parameters now raise a clear `ValueError` instead of crashing with `AttributeError`, and merge-conflicted files are correctly excluded from `modified_files` in `project_status` and handled without crashing in `get_diffs`. Also surfaces all CLI errors as structured JSON when `--json` is set.

## Motivation

In v0.6.x, functions with untyped parameters (e.g. `def f(conv, booking_ref)`) caused `_extract_decorators` to crash with `AttributeError: 'NoneType' has no attribute 'id'`. Separately, files with merge conflicts were being passed to `read_local_resource`, which would now raise rather than silently fail. Both issues needed to be fixed together for `poly status` and `poly diff` to be reliable after a branch pull with conflicts.

Closes #

## Changes

- `_extract_decorators`: guard `arg.annotation` access before reading `.id`; raise `ValueError` with a distinct message for missing vs unsupported type annotations; remove the `try/except SyntaxError` wrapper so errors propagate to the user
- `resource.py`: rename `get_status` → `is_modified`, removing merge conflict detection from the resource itself
- `project_status`: check for merge conflicts on raw file content before calling `read_local_resource`; conflicted files now appear only in `files_with_conflicts`, not also in `modified_files`
- `get_diffs`: same conflict-before-parse pattern; shows diff of conflict markers vs original; fix `type(resource_type)` key bug and missing second arg to `get_diff`
- `cli.py`: wrap the command dispatch in `try/except`; when `--json` is set, errors are returned as `{"success": false, "error": "...", "traceback": "..."}` instead of raising
- `src/poly/docs/functions.md`: note that all parameters must have a supported type annotation

## Test strategy

## Checklist

- [x] `pytest` passes

### DIFF
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -21,6 +21,7 @@ from typing import Any, Optional
 import argcomplete
 import requests
 import questionary
+import traceback
 
 from poly.output.console import (
     console,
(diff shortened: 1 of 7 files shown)


bill-parker (approved):
lgtm!

PR #69, opened 2026-04-10
feat: non-interactive scripted input, conversation resume, pre-chat push, and JSON output for `poly chat`

Ruari-Phipps:
## Summary

Extends `poly chat` with four new capabilities: push before chatting (`--push`), scripted/file-based message input (`-m`/`--input-file`), resuming an existing conversation (`--conv-id`), and structured JSON output (`--json`). Interactive mode is fully unchanged.

## Motivation

Makes `poly chat` usable in automated testing pipelines and CI scripts without a human at the terminal. `--push` removes the manual push step before testing a branch.

## Changes

- `--push`: pushes the project before creating the chat session so local changes are live before testing
- `--message`/`-m` (repeatable): supply utterances non-interactively, e.g. `-m "Hello" -m "Goodbye"`
- `--input-file FILE`: reads messages line-by-line from a file; use `-` for stdin
- `--conv-id`: resumes an existing conversation by ID, skipping session creation entirely
- `--json`: emits a single JSON object `{"conversations": [...]}` when the session ends; each conversation contains `conversation_id`, `url`, and `turns` (greeting is `turns[0]` with `"input": null`); restarts produce multiple entries in the array
- `_run_chat_loop` now returns `(restart, conversation_dict)` so the caller accumulates conversations across restarts before printing
- Clean error message when `--input-file` path does not exist
- Updated `chat_parser` help text with examples for all new flags

## Test strategy

## Checklist

- [ ] `pytest` passes
- [ ] No breaking changes to the `poly` CLI interface (or migration path documented)


src/poly/cli.py
```diff
@@ -17,6 +17,7 @@
 from contextlib import nullcontext
 from importlib.metadata import version as get_package_version
 from typing import Any, Optional
+from xml.parsers.expat import errors
```
alanacevedo: I don't believe this is used in the file, we could remove it

src/poly/cli.py
```diff
@@ -1980,12 +2037,49 @@ def chat(
...
+            if push_success and not output_json:
+                success("Project pushed successfully.")
+
+            if not push_success:
+                if output_json:
+                    json_print(
+                        {
+                            "success": False,
+                            "message": "Failed to push project before chat session.",
+                            "error": error,
```
alanacevedo: I believe error here is the logger func imported from poly.output.console

alanacevedo (approved):

PR #82, opened 2026-04-16
feat: Add 'studio' region and filter region selection based on permissions

milesapnash:
## Summary

Updates `poly init` to only display regions the user's API key has access to, and adds the `studio` region.

## Motivation

Previously, `poly init` showed all hardcoded regions regardless of whether the user had access. This change probes regions concurrently and filters the list. Additionally, `studio` was not available as a region.

## Changes

- Added `get_accessible_regions()` to `PlatformAPIHandler` that concurrently probes regions via `get_accounts()` and returns only accessible ones
- Added `get_accessible_regions()` to `AgentStudioInterface` as the public interface
- Updated `init_project()` in `cli.py` to fetch and display only accessible regions (with a loading spinner), with an error message if none are found
- Added `"studio"` region pointing to `https://api.studio.poly.ai/adk/v1`

## Test strategy

## Checklist

- [x] `pytest` passes

## Screenshots / Logs
Screenshot 2026-04-21 at 11 20 47


src/poly/cli.py
```diff
@@ -1059,13 +1059,24 @@ def init_project(
...
+                    json_print(
+                        {
+                            "success": False,
+                            "error": "No accessible regions found for your API key.",
+                        }
+                    )
+                else:
+                    error("No accessible regions found for your API key.")
+                sys.exit(1)
+            region = questionary.select("Select Region", choices=regions).ask()
```
Ruari-Phipps: If only one region, we should auto select it

milesapnash: Agreed - addressed in [feat: auto-select region when only one is accessible](https://github.com/polyai/adk/pull/82/commits/eb36d15208e656082c52b9790f408ba4a95e6458)

Ruari-Phipps (approved):

PR #94, opened 2026-04-21
feat: Clean KB topic names

Ruari-Phipps:
## Summary
Topic filenames now use `clean_name()` (e.g. `topics/topic_1.yaml`) with the real name stored inside the YAML, matching the flow step pattern. This prevents topics with `/` in their name from being created as nested directories. Existing projects are auto-migrated on load.

## Motivation

Topic names containing `/` (e.g. `"Billing/Refunds"`) were being used directly as filenames, causing `os.path.join("topics", "Billing/Refunds.yaml")` to create nested folder structures instead of a single file.

## Changes

- **`topic.py`**: `file_path` uses `clean_name()`, `to_yaml_dict()` includes `name` field, new `read_local_resource()` override reads name from YAML and validates it matches the filename (with legacy fallback for old-format files)
- **`project.py`**: Added `Topic` to the `find_new_kept_deleted` block that reads resource names from YAML content for newly discovered files; added `_migration_flags` field to track applied migrations, persisted via `to_dict`/`from_dict`/`from_status_dict`; fresh pulls start with all flags set (no migration needed)
- **`migration_utils.py`** (new): Migration framework with `MigrationFlag` enum, `run_migrations()` dispatcher, and `migrate_legacy_topic_files()` — renames old `Topic Name.yaml` files to `topic_name.yaml` and injects the `name` key into the YAML content
- **`migration_test.py`** (new): Tests for the legacy topic file migration — rename, skip already-migrated, in-place update for clean names, mixed files
- **Test fixtures**: Renamed topic fixtures to clean names, added `name` field, updated all path assertions

## Test strategy

## Checklist

- [x] `pytest` passes

### DIFF
--- a/src/poly/utils.py
+++ b/src/poly/utils.py
@@ -12,9 +12,7 @@ import os
 import re
 from typing import Callable, Optional
 
-from poly.resources.flows import FunctionStep
-from poly.resources.function import Function
-from poly.resources.resource import Resource, ResourceMapping
+from poly.resources import resource_utils, Resource, ResourceMapping, Function, FunctionStep, Topic
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


src/poly/utils.py
```diff
@@ -404,3 +402,41 @@ def compute_variable_references(
...
+            continue
+
+        # Read content, insert name field, write to new path, remove old file
+        content = Resource.read_from_file(legacy_path)
+        yaml_dict = resource_utils.load_yaml(content) or {}
+        if "name" not in yaml_dict:
+            yaml_dict = {"name": topic.name, **yaml_dict}
+            new_content = resource_utils.dump_yaml(yaml_dict)
+            with open(new_path, "w", encoding="utf-8") as f:
+                f.write(new_content)
```
milesapnash: "Topic 1" and "topic_1" are currently both potentially legitimate distinct topic names (poor agent design aside), and both would point to the same file in this case, potentially leading to overwrites

Ruari-Phipps: In the ADK, we don't allow if two resources map to the same file path. So I'll maybe flag the error here too

AaronForinton (approved):
approving from my end

milesapnash (approved):

PR #110, opened 2026-04-27
fix: show clear error when POLY_ADK_KEY is not set

oeisenberg:
## Summary
- Running `poly init` without `POLY_ADK_KEY` set now shows `POLY_ADK_KEY environment variable is not set. Export your API key with: export POLY_ADK_KEY=<your-api-key>` instead of the misleading "No accessible regions found for your API key."
- Adds early env var check in `init_project` before any API calls
- Fixes `_retrieve_api_key` in both `PlatformAPIHandler` and `SyncClientHandler` to raise immediately when the key is missing (previously `os.getenv()` silently returned `None`)

## Test plan
- [x] Lint passes (`ruff check`)
- [x] Format passes (`ruff format --check`)
- [x] All 500 tests pass
- [x] Manual: run `unset POLY_ADK_KEY && poly init` and verify the new error message
image

### DIFF
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -50,7 +50,7 @@ from poly.handlers.interface import (
     REGIONS,
     AgentStudioInterface,
 )
-from poly.utils import merge_strings
+from poly.utils import merge_strings, retrieve_api_key
 from poly.resources.resource_utils import contains_merge_conflict
 from poly.project import (
     PROJECT_CONFIG_FILE,
@@ -1247,6 +1247,15 @@ class AgentStudioCLI:
                     "error": "init with --json requires --region, --account_id, and --project_id.",
                 }
             )
+            sys.exit(1)
+
+        try:
+            retrieve_api_key()
+        except ValueError as e:
+            if output_json:
+                json_print({"success": False, "error": str(e)})
+            else:
+                error(str(e))
             sys.exit(1)
 
         if not output_json:
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -1249,15 +1249,6 @@ class AgentStudioCLI:
             )
             sys.exit(1)
 
-        try:
-            retrieve_api_key()
-        except ValueError as e:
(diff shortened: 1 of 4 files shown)


src/poly/cli.py
```diff
@@ -1249,6 +1249,15 @@ def init_project(
...
 
+        try:
+            retrieve_api_key()
+        except ValueError as e:
+            if output_json:
+                json_print({"success": False, "error": str(e)})
+            else:
+                error(str(e))
+            sys.exit(1)
+
```
oeisenberg: get_accessible_regions has a except Exception that silently swallows errors per-region. Without the early check, the ValueError from retrieve_api_key() would be caught there, and you'd still get the misleading "No accessible regions" message.

src/poly/handlers/platform_api.py
```diff
@@ -138,6 +136,10 @@ def get_accessible_regions(regions: list[str]) -> list[str]:
             list[str]: Regions that returned at least one account, preserving
                 the original ordering.
         """
+        # Fail fast if the API key is not configured — don't let the
+        # per-region except swallow the ValueError.
+        PlatformAPIHandler._retrieve_api_key()
```
oeisenberg: Even if someone calls that method outside of init_project, they'll still get the right error

src/poly/handlers/platform_api.py
```diff
@@ -54,11 +53,10 @@ def get_base_url(region: str) -> str:
...
     def _retrieve_api_key() -> str:
-        """Get API key from environment"""
-        try:
-            return os.getenv("POLY_ADK_KEY")
-        except Exception:
-            raise ValueError("POLY_ADK_KEY environment variable is required")
+        """Get API key from environment."""
+        from poly.utils import retrieve_api_key
+
+        return retrieve_api_key()
```
Ruari-Phipps: It should just call this function directly

Ruari-Phipps (approved):

PR #116, opened 2026-04-28
fix: redact API key from debug log output

HelenGreul:
## Summary
- `PlatformAPIHandler.make_request()` logged the full `headers` dict at DEBUG level, which includes the `X-API-KEY` value
- When DEBUG logging is enabled, API keys are written to log files hence potential  credential leak
- Replaced `headers={headers!r}` with `correlation_id={correlation_id!r}` in both the success and error debug log lines, which preserves request traceability without exposing secrets

### DIFF
--- a/src/poly/handlers/platform_api.py
+++ b/src/poly/handlers/platform_api.py
@@ -102,7 +102,7 @@ class PlatformAPIHandler:
         )
 
         logger.debug(
-            f"Request/response url={url!r} headers={headers!r} body={data!r}"
+            f"Request/response url={url!r} correlation_id={correlation_id!r} body={data!r}"
             f" status_code={api_response.status_code!r} response={api_response.text!r}"
         )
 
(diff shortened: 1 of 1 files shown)


HelenGreul:
CI failure seems unrelated to this change - all test and coverage steps passed. The _post coverage comment_ step fails because the GH Actions token doesn't have write permissions on fork PRs

src/poly/handlers/platform_api.py
```diff
@@ -102,15 +102,15 @@ def make_request(
...
+            f"Request/response url={url!r} correlation_id={correlation_id!r} body={data!r}"
             f" status_code={api_response.status_code!r} response={api_response.text!r}"
         )
 
         try:
             api_response.raise_for_status()
         except requests.HTTPError:
             logger.debug(
-                f"Error in request. url={url!r} headers={headers!r} body={data!r}"
+                f"Error in request. url={url!r} correlation_id={correlation_id!r} body={data!r}"
```
Ruari-Phipps, line 113: In an error we would relog url/id/body, one in initial request and one in error. Might be worth only logging the top one on success

Ruari-Phipps:
Merged a change that should fix the coverage report issue, if you merge/rebase it _should_ work now but I think you can merge without it

PR #121, opened 2026-04-30
fix: cross-platform path handling + add Windows CI

alanacevedo:
## Summary
- Fixes cross-platform bugs in production code where `file_path.split(os.sep)` failed on Windows because paths containing forward slashes weren't normalized first — adds `os.path.normpath()` before splitting
- Normalizes topic names in `migration_utils.py` to always use `/` regardless of OS
- Adds a `test-windows` CI job (`continue-on-error: true`) to catch platform-specific bugs
- Updates test assertions to use `os.path.join()` instead of hardcoded forward-slash paths

## Production code changes
| File | Fix |
|---|---|
| `src/poly/resources/function.py` | `os.path.normpath()` before `split(os.sep)` in `get_function_type()` and `read_local_resource()` |
| `src/poly/resources/flows.py` | Same in `FlowStep` condition matching and `FunctionStep.read_local_resource()` |
| `src/poly/resources/resource_utils.py` | Same in `get_flow_name_from_path()` |
| `src/poly/migration_utils.py` | `os.path.relpath` returns OS-native separators; normalize to `/` for topic names |

## Test changes
| File | Fix |
|---|---|
| `src/poly/tests/project_test.py` | `os.path.join()` in `get_diffs` assertions |
| `src/poly/tests/resources_test.py` | `os.path.join()` for mock dict keys and file_path args; skip Unix-specific path test on Windows |

## Test plan
- [x] CI passes on Ubuntu, Windows, and coverage

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,8 +32,11 @@ jobs:
         fi
         echo "PR title is valid: $PR_TITLE"
 
-  lint-and-test:
-    runs-on: ubuntu-latest
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
     env:
       CI: true
     permissions:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -88,3 +88,35 @@ jobs:
...
+        python-version: "3.14"
+
+    - name: Create virtual environment
+      run: uv venv
+
+    - name: Install dependencies
+      run: uv pip install -e ".[dev]"
+
+    - name: Run tests
+      run: uv run pytest src/poly/tests/ -v
```
Ruari-Phipps: Can we have this in the lint-and-test using a matrix? Thats what claude suggested to me

alanacevedo: Done — switched to a matrix on `lint-and-test` with `[ubuntu-latest, windows-latest]`. Lint/format/license steps are gated with `if: matrix.os == 'ubuntu-latest'`. Also replaced `source .venv/bin/activate` with `uv run` since activate doesn't work on Windows.

Ruari-Phipps (approved):
LGTM!

PR #123, opened 2026-05-01
feat: support per-region API keys via POLY_ADK_KEY_{REGION} for the incoming PAT change

oeisenberg:
## Summary

- `retrieve_api_key()` now accepts an optional `account_id` parameter
- When provided, tries `POLY_ADK_KEY_{SUFFIX}` first (e.g. `POLY_ADK_KEY_US`), then falls back to `POLY_ADK_KEY`
- This adds support for PAT which are incoming.

## Test plan

- [x] Lint passes (`ruff check`)
- [x] Format passes (`ruff format --check`)
- [x] All 552 tests pass
- [x] Manual: set `POLY_ADK_KEY_US=<key>` and run `poly init`
- [x] Manual: update POLY_ADK_KEY_US=<key>` to fail and set `POLY_ADK_KEY` — should fall back gracefully
- [x] Manual: unset both — should show the existing error message

### DIFF
--- a/src/poly/utils.py
+++ b/src/poly/utils.py
@@ -21,24 +21,6 @@ _TYPES_PACKAGE = "poly.types"
 _API_KEY_ENV_VAR = "POLY_ADK_KEY"
 
 
-def _region_to_env_suffix(region: str) -> str:
-    """Derive env-var suffix from a region identifier.
-
-    Strips a trailing ``-<digit>`` segment (e.g. ``-1``) and uppercases.
-
-    Examples:
-        us-1    -> US
-        euw-1   -> EUW
-        uk-1    -> UK
-        studio  -> STUDIO
-        staging -> STAGING
-        dev     -> DEV
-    """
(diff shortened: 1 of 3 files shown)


src/poly/utils.py
```diff
@@ -21,8 +21,21 @@
...
+      1. ``POLY_ADK_KEY_{REGION}`` (if *region* is provided, e.g. ``POLY_ADK_KEY_US``)
+      2. ``POLY_ADK_KEY``
+
+    Raises ``ValueError`` with a helpful message when no key is found.
+    """
+    if region:
+        suffix = re.sub(r"-\d+$", "", region).upper()
+        per_region_key = os.getenv(f"{_API_KEY_ENV_VAR}_{suffix}")
+        if per_region_key:
+            return per_region_key
```
Ruari-Phipps: Maybe just a dictionary mapping rather than constructing it with a regex?

Ruari-Phipps (approved):

PR #148, opened 2026-05-15
fix: Remove project_id prompt for Studio region

Ruari-Phipps:
## Summary

Removes the default project_id slug generation and skips the project_id prompt entirely for Studio (`--region studio`), since project_id must be unique across all projects in the region and prompting for it in PLG causes friction.

## Motivation

- `project_id` must be globally unique per region — if the user picks a taken ID, the API errors. We can't validate uniqueness client-side.
- For PLG (Studio) we expect many projects, so defaulting to empty and letting the platform generate the ID avoids collisions.
- The default slugified name was often not useful and led to conflicts.

## Changes

- Remove default slug generation from `project_name` for the `project_id` field
- Skip the `project_id` interactive prompt when `--region studio`
- Update help text to clarify the platform generates the ID if omitted
- Fix `is not` → `!=` for string comparison (identity vs equality bug)

## Test strategy

- [x] `pytest` passes

## Checklist

- [x] `pytest` passes

### DIFF
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -1534,7 +1534,7 @@ class AgentStudioCLI:
                 return
             project_name = project_name.strip()
 
-        if not project_id and region is not "studio":
+        if not project_id and region != "studio":
             project_id = questionary.text(
                 "Enter project ID (leave empty to let the platform generate one):",
                 validate=lambda val: (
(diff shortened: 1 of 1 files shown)


milesapnash (approved):

bplevin36 (approved):
good change

PR #153, opened 2026-05-18
fix: Correct merge_branch return type annotations in project.py

Ruari-Phipps:
## Summary

Fixes the return type annotation on `AgentStudioProject.merge_branch()` to match the actual return type from the handler layer.

## Motivation

The `project.py` method declared `list[str]` for the conflicts and errors return values, but the handlers (`interface.py`, `sync_client.py`) correctly return `list[dict[str, str]]`. This mismatch could mislead type checkers and developers.

## Changes

- Updated `merge_branch` return type from `tuple[bool, list[str], list[str]]` to `tuple[bool, list[dict[str, str]], list[dict[str, str]]]`
- Updated docstring to match

## Test strategy

- [x] N/A (docs, config, or trivial change)

## Checklist

- [x] `pytest` passes

### DIFF
--- a/src/poly/project.py
+++ b/src/poly/project.py
@@ -2700,7 +2700,7 @@ class AgentStudioProject:
 
     def merge_branch(
         self, message: str, conflict_resolutions: list[dict[str, Any]] = None
-    ) -> tuple[bool, list[str], list[str]]:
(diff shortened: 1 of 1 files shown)


oeisenberg (approved):
lgtm

PR #156, opened 2026-05-19
refactor: Replace email param threading with POLY_ADK_EMAIL env var

Ruari-Phipps:
## Summary

Replace the `email` parameter that was threaded through the entire call chain (CLI -> project -> interface -> sync_client -> SDK) with a single `ADK_COMMAND_USER_OVERRIDE` environment variable read once in `SourcererSDK.__init__`.

## Motivation

The email was being passed as a parameter through 5 layers of function calls just to set a header and metadata field. Using an env var simplifies the API, removes parameter plumbing, and ensures the email header is consistently included on all API requests (including merges, which previously didn't get it).

## Changes

- Read `ADK_COMMAND_USER_OVERRIDE` env var in `SourcererSDK.__init__`, set on session headers and `send_command_batch` headers
- Use `self.email` as default `created_by` in `create_metadata()`
- Remove `email` param from `queue_resources`, `send_queued_commands`, `send_command_batch`, `_stage_commands`, `push_project`, `sync_ids_with_sandbox`
- Remove `--email` CLI flag from push command
- Remove `email` param from `AgentStudioInterface.queue_resources` and `upload_resources`

## Test strategy

- [x] N/A (docs, config, or trivial change)

## Checklist

- [ ] `pytest` passes

### DIFF
--- a/src/poly/handlers/sdk.py
+++ b/src/poly/handlers/sdk.py
@@ -80,7 +80,7 @@ class SourcererSDK:
 
         # Initialize session with auth headers
         correlation_id = f"adk-{uuid.uuid4()}"
-        self.email = os.environ.get("POLY_ADK_EMAIL")
+        self.email = os.environ.get("ADK_COMMAND_USER_OVERRIDE")
         self.session = requests.Session()
         headers = {
             "Content-Type": "application/json",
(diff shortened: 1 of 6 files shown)


src/poly/handlers/sdk.py
```diff
@@ -80,14 +80,16 @@ def __init__(
 
         # Initialize session with auth headers
         correlation_id = f"adk-{uuid.uuid4()}"
+        self.email = os.environ.get("ADK_COMMAND_USER_OVERRIDE")
```
bplevin36, line 83: is there any doc anywhere that it would make sense to mention this? if not, please do put a comment here explaining what the side effect of setting this header is

bplevin36 (approved):
just one nit to address

PR #162, opened 2026-05-27
feat: support conditions on function steps

Ruari-Phipps:
## Summary

Adds full sync support for flow conditions on function steps, including pull, push, YAML serialization, and the correct Agent Studio protobuf APIs (`CreateStepCondition` / `UpdateStepCondition` / `DeleteStepCondition` instead of the no-code condition endpoints).

## Motivation

Function steps can define branching via `goto_step()` in their Python code, but the ADK only handled conditions for no-code steps. Conditions on function steps were dropped on pull and could not be pushed reliably. The backend also auto-creates conditions from `goto_step()` calls, which caused duplicate labels when pushing explicit conditions alongside real function code.

## Changes

- Add `conditions` field to `FunctionStep` with YAML read/write and `goto_step()` extraction (mirroring `FlowStep`)
- Route function-step conditions through `step_condition` protos; no-code step conditions unchanged
- Track `parent_is_no_code_step` on `Condition` to select the correct create/update/delete API
- Two-phase push for new function steps with conditions: create with stub `pass` code first, then update with real code to avoid backend duplicate conditions
- Include `Condition` in sync ordering so conditions are created before dependent updates
- Pull function-step conditions from the platform API in `sync_client`
- Add unit tests for pull, push, and YAML round-trip of function-step conditions

## Test strategy

## Checklist

- [ ] `pytest` passes

## Screenshots / Logs

N/A

Made with [Cursor](https://cursor.com)


Ruari-Phipps:
Still active

PR #177, opened 2026-06-09
fix: persist new test case prompt_assertions and tags on push

arkadoosh:
## Summary

A newly-created `TestCase` was pushed with its base fields only — its `prompt_assertions` and `tags` were silently dropped, so a `push` → `pull` round-trip lost them (the case came back with a `scenario` but no `prompt_assertions`).

## Motivation

Surfaced in Glot, which authors `test_suite/*.yaml` cases, pushes, then runs them: after the projection round-trip the cases came back assertion-less, so the runner skipped them. The cases only "stuck" on a *second* push, once the case already existed.

Root cause: `TestCase.get_new_updated_deleted_subresources` returns its assertions and tags in the **`updated`** bucket even for a brand-new case — they have no create proto and are only settable via `set_test_case_assertions` / `set_test_case_tags`. But `AgentStudioProject._get_updated_subresources` only forwarded the **`new`** bucket for newly-created resources, so the `set_test_case_assertions` / `set_test_case_tags` commands were never emitted. The case landed in the projection with no assertions; pull (`_read_test_cases_from_projection`) faithfully reconstructed an assertion-less case.

## Changes

- `_get_updated_subresources`: for newly-created resources, forward the `updated` and `deleted` sub-resources too, not just `new`. Creates are already ordered before updates, so `Create_TestCase` is followed by `set_test_case_assertions` / `set_test_case_tags`.
- This also fixes the same latent drop for a new `Function`'s non-default `latency_control`, which is bucketed identically.
- Adds `GetUpdatedSubresourcesTest.test_new_test_case_emits_assertions_and_tags` asserting a brand-new `TestCase` emits its assertions/tags in the `updated` change-set.

## Test strategy

> Couldn't run `pytest` locally (a `ruamel` namespace-package quirk in my env); the new test targets the exact buckets and CI exercises it.

## Checklist

- [x] `pytest` passes _(verified via CI — see note above)_

## Screenshots / Logs

Before: a freshly-authored `test_suite/<name>.yaml` round-trips (push→pull) with its `scenario` but **no `prompt_assertions`**.
After: assertions and tags survive the round-trip.

### DIFF
--- a/src/poly/project.py
+++ b/src/poly/project.py
@@ -1324,6 +1324,17 @@ class AgentStudioProject:
                     new_subresources.setdefault(type(sub_resource), {})[
                         sub_resource.resource_id
                     ] = sub_resource
+                # A new parent can carry update-only sub-resources — e.g. a TestCase's
+                # prompt_assertions/tags, applied via set_test_case_assertions (no create
+                # proto). Forward updated + deleted too, or they're dropped on create.
+                for sub_resource in updated:
+                    updated_subresources.setdefault(type(sub_resource), {})[
+                        sub_resource.resource_id
+                    ] = sub_resource
+                for sub_resource in deleted:
+                    deleted_subresources.setdefault(type(sub_resource), {})[
+                        sub_resource.resource_id
+                    ] = sub_resource
 
         return SubResourceChangeSet(
             new=new_subresources,
--- a/src/poly/tests/project_test.py
+++ b/src/poly/tests/project_test.py
@@ -34,6 +34,8 @@ from poly.resources import (
     SMSTemplate,
     Topic,
     TestCase,
+    TestCaseAssertion,
+    TestCaseTags,
     TranscriptCorrection,
     Translation,
     Variable,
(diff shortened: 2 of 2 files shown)


src/poly/project.py
```diff
@@ -1324,6 +1324,17 @@ def _get_updated_subresources(
...
+                # via an update command, not a create one — e.g. a TestCase's
+                # prompt_assertions/tags, whose sub-resource has no create proto and is
+                # applied with `set_test_case_assertions`. These come back in the
+                # `updated` bucket; without emitting them here a brand-new test case is
+                # pushed with no assertions/tags (they only persisted on a later update),
+                # so push -> pull silently drops `prompt_assertions`.
+                for sub_resource in updated:
+                    updated_subresources.setdefault(type(sub_resource), {})[
+                        sub_resource.resource_id
+                    ] = sub_resource
```
Ruari-Phipps:
```suggestion
                for sub_resource in updated:
                    updated_subresources.setdefault(type(sub_resource), {})[
                        sub_resource.resource_id
                    ] = sub_resource
                for sub_resource in deleted:
                    deleted_subresources.setdefault(type(sub_resource), {})[
                        sub_resource.resource_id
                    ] = sub_resource
```

arkadoosh: Good call — applied. The new-resources branch now forwards both `updated` and `deleted` sub-resources (`deleted` is empty for a brand-new parent, but kept for symmetry with the updated-resources path). Pushed in e13795c.

Ruari-Phipps (approved):
LGTM! Thanks!

PR #189, opened 2026-06-15
refactor: rewrite stub sync script to use mypy stubgen

Ruari-Phipps:
## Summary

Replace the custom AST-based stub generator with mypy's `stubgen` plus a post-processing pass that emits runtime-safe `.py` stub modules — the committed files are byte-identical to what `poly pull` copies into a project's `_gen/`, so review and CI validate exactly what ships. Regenerated from the latest runtime.

## Motivation

The custom script (~600 lines) reimplemented stub generation with edge cases around dataclass `__init__` synthesis, import pruning, `TYPE_CHECKING` blocks, and depth-aware relative imports. `stubgen` handles all of this correctly out of the box. Regenerating also picks up the runtime modules added since the stubs were last synced.

## Changes

- Rewrite `scripts/sync_runtime_stubs.py` to shell out to `stubgen` (via `uv run --extra dev`) and post-process into runtime-safe `.py` modules: relativize `runtime.`/`utils.` imports, inject `__all__` from the runtime's `imports.json`, drop imports of internal-only modules and erase their names to `Any`, bind annotation-only declarations (`X: T = ...`), add `from __future__ import annotations` and the copyright + linter-suppression header
- Add a typing-closure list (`CLOSURE_SOURCES`): modules such as `utils/api_connector.py` are stubbed so that exported classes keep their real base classes and annotation types, but get no `__all__` entry, so nothing advertises them as importable (their back-edge imports are guarded under `TYPE_CHECKING` so the package has no import cycles); re-exports of private names stubgen omits are dropped
- Validate at sync time: the script refuses to run unless the runtime checkout is a clean `origin/main` (`--skip-git-check` to override), imports every generated module in a subprocess and asserts every `__all__` name binds, then runs `ruff check --ignore-noqa --select F821,F822` and `ruff format` over the output
- Write `src/poly/types/_manifest.json` (runtime commit, generation time, module count) so the shipped stubs' origin is recorded and verifiable
- Simplify `save_imports()` to a verbatim copy of the committed modules plus the two genuinely per-project files (aggregated `__init__.py` with relative imports, `decorators.py`)
- Add `scripts/verify_wheel.py` + a `wheel-check` CI job: builds the wheel from a clean source copy (a stale `build/` tree would otherwise mask packaging regressions), installs it into a fresh venv, and generates a `_gen/` from it — the editable installs used everywhere else cannot catch packaging mistakes
- Regenerate all stubs from current runtime `main`, adding modules that were missing (`clock`, `connectors/`, `cxone`, `janus_sms`, `outbound`)
- Add tests: generate a real `_gen/`, import it, and assert every `__all__` name is bound; import-walk the committed `poly.types` modules directly; assert no stub carries absolute `runtime.`/`utils.` imports; check the provenance line matches the manifest

Usage: `python scripts/sync_runtime_stubs.py [--python-root /path/to/genai_lambda_runtime/python]`

## Test strategy

The sync script was also run end-to-end against a fresh runtime checkout, and `save_imports()` was verified against a real Agent Studio project: the regenerated `_gen/` passes `ruff` and imports cleanly with every exported name bound.

## Checklist

- [x] `pytest` passes (2059 tests, 434 subtests)


Ruari-Phipps:
Still active

PR #191, opened 2026-06-15
fix(project): hint at unquoted-YAML cause on resource-read TypeError

arkadoosh:
## Summary

Make the opaque `/push` failure on malformed resource YAML actionable: `read_local_resource` now catches `TypeError` specifically and re-raises with the file path it already reports **plus** the likely cause (an unquoted YAML scalar parsed as the wrong type).

## Motivation

When a pushed resource YAML has an unquoted scalar that parses as the wrong type — a string with a mid-sentence colon (`outcome: either ...`) read as a mapping, or a leading `[ { * & ?` read as a sequence/anchor — `push_project` raises a bare CPython `TypeError` ("bad argument type for built-in operation"). The `/push` HTTP service surfaces that verbatim as a 422, so consumers (e.g. Glot, the Agent Studio assistant) get an error that names no file or field and bisect blindly to find the offending value.

Linear: DEVP-331

## Changes

- `read_local_resource` (`src/poly/project.py`): add a `TypeError` branch that re-raises a `ValueError` keeping the file path + original message and appending the likely cause (unquoted YAML value parsed as the wrong type — quote it). Non-`TypeError` failures keep the existing plain wrapping, so no spurious hint is added to unrelated errors.

## Test strategy

`ReadLocalResourceErrorTest` in `src/poly/tests/project_test.py`: a `TypeError` from the resource read is wrapped with the file path + YAML hint; a non-`TypeError` keeps the plain message with no hint.

## Checklist

- [x] `pytest` passes (new tests + nearby read/discover/find suites)

## Screenshots / Logs

Before: `ADK /push returned status 422: {"detail":"bad argument type for built-in operation"}`
After: `Error reading resource <name> at <file>: bad argument type for built-in operation. A YAML value is likely unquoted and parsed as the wrong type — check for an unquoted mid-sentence colon or a leading [ { * & ? and quote the value.`

### DIFF
--- a/src/poly/project.py
+++ b/src/poly/project.py
@@ -2104,18 +2104,6 @@ class AgentStudioProject:
             raise FileNotFoundError(
                 f"File not found for resource {resource.resource_name} at {resource.file_path}"
             ) from e
-        except TypeError as e:
-            # A TypeError here almost always means a YAML value parsed as the
-            # wrong type: an unquoted string with a mid-sentence colon parses as a
-            # mapping, a leading [ { * & ? as a sequence/anchor/alias. The raw
-            # CPython message (e.g. "bad argument type for built-in operation")
-            # names no field, so add the likely cause.
-            raise ValueError(
-                f"Error reading resource {resource.resource_name} at "
-                f"{resource.file_path}: {str(e)}. A YAML value is likely unquoted "
-                f"and parsed as the wrong type — check for an unquoted mid-sentence "
-                f"colon or a leading [ {{ * & ? and quote the value."
-            ) from e
         except Exception as e:
             raise ValueError(
                 f"Error reading resource {resource.resource_name} at {resource.file_path}: {str(e)}"
--- a/src/poly/project.py
+++ b/src/poly/project.py
@@ -2104,6 +2104,18 @@ class AgentStudioProject:
             raise FileNotFoundError(
                 f"File not found for resource {resource.resource_name} at {resource.file_path}"
             ) from e
+        except TypeError as e:
+            # A TypeError here almost always means a YAML value parsed as the
+            # wrong type: an unquoted string with a mid-sentence colon parses as a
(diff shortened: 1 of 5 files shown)


src/poly/project.py
```diff
@@ -2104,6 +2104,18 @@ def read_local_resource(
...
+            # wrong type: an unquoted string with a mid-sentence colon parses as a
+            # mapping, a leading [ { * & ? as a sequence/anchor/alias. The raw
+            # CPython message (e.g. "bad argument type for built-in operation")
+            # names no field, so add the likely cause.
+            raise ValueError(
+                f"Error reading resource {resource.resource_name} at "
+                f"{resource.file_path}: {str(e)}. A YAML value is likely unquoted "
+                f"and parsed as the wrong type — check for an unquoted mid-sentence "
+                f"colon or a leading [ {{ * & ? and quote the value."
+            ) from e
```
Ruari-Phipps: Can we include this within the `load_yaml` function rather than here?

arkadoosh: Done in 8239a75 — `load_yaml` now catches the parse failure and re-raises a `ValueError` with the hint, so every caller benefits and the special-case is gone from `read_local_resource`.

One tradeoff worth flagging: the original opaque error has two shapes. A value that *fails to parse* (mid-sentence colon, unclosed `[ {`, undefined alias) raises inside `load_yaml`, so that now gets the hint everywhere. But a value that parses as a *valid* collection (`[a, b]`, `{a: b}`) only blows up later in `from_yaml_dict` with the bare `TypeError: bad argument type for built-in operation` — that is downstream of `load_yaml`, so it now falls through to the generic `Error reading resource ...` wrapper without the hint. The common footgun is covered; the rarer wrong-type-but-valid case is not. Happy to add a `TypeError` guard back at the construction site if you want that covered too.

Tests now trigger the real ruamel error instead of mocking a `TypeError`.

PR #225, opened 2026-07-09
fix: normalize personality adjectives on pull and send removals on push

Ruari-Phipps:
## Summary

Fix two bugs in personality adjective handling that caused phantom diffs on pull and silent no-ops when removing adjectives.

## Motivation

1. Pulling a project could produce spurious diffs because the platform may return `{Friendly: false}` or `{}` for the same state — both mean "not enabled" but produced different YAML.
2. Removing an adjective from the local YAML and pushing had no effect because `build_update_proto` only sent adjectives present in the local dict, never signalling a removal.

## Changes

- `to_yaml_dict` now filters out disabled (`false`) adjectives so pull always writes the same YAML regardless of API representation
- `build_update_proto` now iterates all `ALLOWED_ADJECTIVES`, defaulting missing ones to `false`, so removals are explicitly sent to the platform
- Updated and added tests for both behaviors

## Test strategy

## Checklist

- [x] `pytest` passes

### DIFF
--- a/src/poly/resources/agent_settings.py
+++ b/src/poly/resources/agent_settings.py
@@ -54,7 +54,10 @@ class SettingsPersonality(YamlResource):
 
     def to_yaml_dict(self) -> dict:
         """Convert the personality settings to a YAML-serializable dict."""
-        return {"adjectives": self.adjectives, "custom": self.custom}
+        return {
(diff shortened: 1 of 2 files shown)


jamesosullivan-polyai (approved):
Awesome

PR #226, opened 2026-07-09
feat: test-case severity + poly test run --severity

Wiz1991:
## Summary

Adds support for the platform's new **test-case severity** — a single level per test (`low | medium | high | critical`, or none) — and a `--severity` selector for `poly test run`. Reworked from the earlier labels version after the product pivot (PolyAI-LDN/platform_ui#8394): severity is a scale, one value per test, not a multi-value grouping.

Depends on the platform severity backend, **merged + deploying** (PolyAI-LDN/platform_ui#8394). Until dev/prod roll out, `--severity`/`--tag` 422 cleanly.

## Changes
- **`test_suite` resource**: `TestCaseSeverity` sub-resource — `severity:` on test YAMLs round-trips pull/push and emits `set_test_case_severity` (including for newly created files); omitting the value clears it; local validation against the scale (server zod enum authoritative).
- **`poly test run --severity`** (choices in `--help`): server-side selection via `{ branchId, select: { mode: 'severities', severities: [...] } }` — multiple levels = run tests at any of them.
- `--tag` remains server-side (`mode: 'tags'`); `--files` local → explicit ids; selectors mutually exclusive; `--dry-run` local with a server-authoritative caveat.
- **Protobuf stub refresh** from current platform protos (protoc 25.5): `SetTestCaseSeverity`/`TestCase.severity`; labels field 17 + command 434 are reserved upstream; picks up `set_topic_tags`; drops orphaned `tts_settings_pb2`.

## Testing
`uv run --extra dev pytest src/poly/tests`: **795 passed** — YAML round-trip incl. clear-severity, scale validation, new-file sub-resource emission, resolve_tests severity filtering (OR, unmatched raises), select-body construction. ruff check + format clean.


src/poly/resources/test_suite.py
```diff
@@ -41,6 +42,10 @@
 
 ALLOWED_TYPES = ["string", "integer", "number", "boolean"]
 
+# Predefined workflow labels; the closed set is enforced server-side
+# (zod enum) — kept in sync here for early local validation and --help.
+ALLOWED_LABELS = ["smoke", "regression", "release-gate", "critical"]
```
Wiz1991: Is this an ok pattern or should we just let validation fail?

Ruari-Phipps: Yeah this is good! We try to ensure nothing can be pushed that will fail the remote validation so being able to validate locally is good

Ruari-Phipps (approved):
Nice! Lgtm!

src/poly/resources/test_suite.py
Wiz1991: only con is that in case we do add new labels, we'll end up having to sync it here

adk will always lag behind?

PR #227, opened 2026-07-09
feat: add branch diff, review, and status commands

Ruari-Phipps:
## Summary

Add `poly branch diff`, `poly branch review`, and `poly branch status` commands that compare a branch against its fork-point state, mirroring the UI's branch comparison screen.

## Motivation

The platform now shows branch comparisons relative to the fork point (not latest main), powered by `?atSequence=` on the projection endpoint. ADK had no CLI equivalent — `poly diff` compares local vs remote, not branch vs fork-point. These commands fill that gap as the "review before merge" workflow alongside `branch merge`.

## Changes

- Add `poly branch diff` — prints fork-point diff to terminal (supports `--files`, `--json`)
- Add `poly branch review` — creates a GitHub Gist of the fork-point diff for sharing
- Add `poly branch status` — shows new/modified/deleted files on a branch vs fork point, same output format as `poly status` with extra branch metadata in the header
- Update `get_branches()` across the stack to return `dict[str, dict[str, Any]]` (full branch metadata including `parentBranchId`, `parentSequence`, `isDiverged`) instead of `dict[str, str]`
- Add `branch_id` and `at_sequence` params to `fetch_projection()` for fetching historical projections without cache mutation
- Add `pull_branch_resources()` to sync client and interface layers
- Extract `_resolve_branch_fork_point()` in project layer, shared by `diff_branch()` and `branch_status()`
- Refactor `StatusCommand.status()` to use shared helpers from `shared.py`
- Fallback: when `parentSequence` is null (older branches), compares against latest parent with a warning

## Test strategy

## Checklist

- [x] `pytest` passes


src/poly/project.py
```diff
@@ -1787,6 +1787,146 @@ def diff_resource_maps(
...
+                does not exist.
+        """
+        _, branches = self.get_branches()
+
+        if branch_name:
+            if branch_name not in branches:
+                raise ValueError(f"Branch '{branch_name}' does not exist.")
+            branch_meta = branches[branch_name]
+        else:
+            current_name, _ = self.get_branches()
```
jamesosullivan-polyai: Nit: looks like we don't need call this twice and it can be consolidated to just 1 call on line 1805

src/poly/cli_commands/branch.py
```diff
@@ -906,3 +967,175 @@ def branch_merge(
...
+            else:
+                error(f"GitHub API error: {e}")
+        except OSError as e:
+            if output_json:
+                json_print({"success": False, "message": str(e)})
+            else:
+                error(str(e))
+
+    @classmethod
+    def branch_status(
```
jamesosullivan-polyai, line 1073: nit: seems like this function is missing tests?

Wiz1991 (approved):

jamesosullivan-polyai (approved):

PR #239, opened 2026-07-28
perf: avoid YAML serialize/reparse roundtrip in multi-resource reads

Ruari-Phipps:
## Summary

Removes a redundant YAML serialize→reparse roundtrip from the resource read path. Multi-resource reads now use the already-parsed cached dict directly, cutting `poly diff` on a large project from ~4.0s to ~1.8s (2.3×, measured with `time poly diff`).

## Motivation

Profiling `poly diff` on a large agent project showed the time dominated by ruamel.yaml (serialize + parse). The cause: `MultiResourceYamlResource` caches each file already parsed as a dict, but on every resource read it dumped the matching sub-dict back to a YAML string just so the string-based name→ID replacement could run, then immediately reparsed that string into a dict. The dict was squeezed through a string and reconstituted for no benefit (the loader is `typ="safe"`, so no comments/formatting are preserved).

## Changes

- `resource_utils.py`: add `replace_resource_names_with_ids_in_data`, a dict/list-walking equivalent of the existing string replacement; factor the shared lookup/replacer into `_build_reference_replacer` (behaviour of the two string functions is unchanged).
- `resource.py`: add a `_read_yaml_dict` read seam. `MultiResourceYamlResource` returns its cached sub-dict via a single `_get_matching` seam that both `read_from_file` (string) and `_read_yaml_dict` (dict) derive from. Move the generic name→ID replacement into the base `from_pretty_dict` (it now walks the dict and returns fresh containers), instead of a whole-string pass in `read_local_resource`.
- `flows.py`, `test_suite.py`, `phrase_filter.py`, `variant_attributes.py`: the four `from_pretty_dict` overrides now call `super().from_pretty_dict()` first so they keep the generic replacement.
- `languages.py`: `DefaultLanguage` / `AdditionalLanguage` override `_get_matching` (their on-disk form is a bare scalar) instead of `read_from_file`, removing the need for a special-case flag.

## Test strategy

- [x] Manual CLI testing (`poly diff` on a large project — output byte-identical, "No changes detected.")

New unit tests assert the dict-walking replacement matches the old string-path result on nested structures, preserves non-string leaves and keys, does not mutate its input, and honours flow-scoped mappings.

## Checklist

- [x] `pytest` passes (946 passed)

## Screenshots / Logs

`time poly diff` on a large project (median of 5 warm runs):

| | Before | After |
|---|---|---|
| Wall time | **3.97s** | **1.75s** |

~2.3× faster; output unchanged (`No changes detected.`).

_(Earlier drafts of this PR quoted cProfile totals, which inflate ruamel's per-call cost; the above are real `time poly diff` wall-clock numbers.)_

### DIFF
--- a/src/poly/resources/languages.py
+++ b/src/poly/resources/languages.py
@@ -119,6 +119,7 @@ class DefaultLanguage(MultiResourceYamlResource):
 
     @classmethod
     def _get_matching(cls, file_path: str) -> dict:
+        """Return the default language as a dict, reshaped from its bare scalar on-disk form."""
         true_file_path, _ = _parse_multi_resource_path(file_path)
         value = cls._get_top_level_data(true_file_path).get("default_language")
         if not value:
@@ -240,6 +241,8 @@ class AdditionalLanguage(MultiResourceYamlResource):
 
     @classmethod
     def _get_matching(cls, file_path: str) -> dict:
+        """Return the named additional language as a dict, reshaped from its bare scalar
+        on-disk form."""
         true_file_path, segments = _parse_multi_resource_path(file_path)
         resource_name = segments[-1]
         additional = cls._get_top_level_data(true_file_path).get("additional_languages") or []
--- a/src/poly/resources/languages.py
+++ b/src/poly/resources/languages.py
@@ -118,13 +118,12 @@ class DefaultLanguage(MultiResourceYamlResource):
         _write_languages(type(self), true_file_path, top_level, save_to_cache)
 
     @classmethod
-    def read_from_file(cls, file_path: str) -> str:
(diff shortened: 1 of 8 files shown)


src/poly/resources/languages.py
```diff
@@ -118,13 +118,12 @@ def save(
         _write_languages(type(self), true_file_path, top_level, save_to_cache)
 
     @classmethod
-    def read_from_file(cls, file_path: str) -> str:
+    def _get_matching(cls, file_path: str) -> dict:
```
jamesosullivan-polyai, line 121: nit: can we add docstrings to each function

jamesosullivan-polyai (approved):
lgtm

PR #245, opened 2026-07-29
feat: add poly functions commands and grouped CLI help

andymohajeri:
## Summary

Adds a `poly functions` command group that wraps the newly published public Functions REST API and covers new functionality not previously represented in the ADK.

- `execute` — run a function by name (or ID) with given arguments, returning its output, logs, and runtime.
- `validate` — check every function on the branch for syntax errors and orphaned flow-step references.

The push/pull Functions workflow already present in the ADK has not been affected and works the same as prior to this PR.

## Motivation

- [DEVP-541: Integrate Functions API into ADK](https://linear.app/poly-ai/issue/DEVP-541/integrate-functions-api-into-adk)
- [DEVP-543: Support grouped section headers in poly CLI --help output](https://linear.app/poly-ai/issue/DEVP-543/support-grouped-section-headers-in-poly-cli-help-output)

DEVP-543 is bundled in at reviewer request. The two changes are independent and are separate commits.

## Changes

- `poly functions execute <function_name>` — run a function, print its return value, logs, and runtime. Accepts `--args` and `--region`/`--project_id`/`--branch_id` for headless use.
- `poly functions validate` — validate all functions on the current branch.
- Grouped `--help` section headers for `poly`, `poly functions`, and `poly branch`.
- Docs: dedicated `docs/docs/reference/cli/functions.md` page; removed duplicate legacy inline command docs from `cli.md`.

## Test strategy

## Checklist

- [x] `pytest` passes (1342 passed)

## Screenshots / Logs

`poly functions --help`:

```
usage: poly functions [-h] [--verbose] <subcommand> ...

Manage Functions via the public Functions REST API, scoped to the
project's current branch. Creating, editing and deleting functions is
still done via the local-file/decorator workflow (poly push/poly pull);
this covers what that workflow can't: running and validating
functions.

Examples:
  poly functions execute <function_name> --args '{"x": 1}'
  poly functions validate

Run and inspect:
    execute            Execute a function with the given arguments.
    validate           Validate all functions on the current branch.

options:
  -h, --help           show this help message and exit
  --verbose            Show full error tracebacks for debugging.
```


src/poly/tests/functions_api_test.py
Ruari-Phipps, line 1: This should go in a separate file

josephcesarperez (approved):
lgtm functionally, just curious about the context for why we are introducing the usecase splitting out from the usual local pull push workflow

docs/docs/reference/cli.md
```diff
@@ -16,6 +16,15 @@ To see all available commands and options:
 poly --help
 ~~~
 
+Commands are listed under section headers so related ones stay together:
+
+| Section | Commands |
+|---|---|
+| Getting started | `init`, `start`, `login`, `studio`, `project` |
+| Project sync | `pull`, `push`, `status`, `revert`, `format`, `validate`, `diff`, `review`, `branch` |
+| Builder API | `deployments`, `conversations`, `audio-cache`, `functions`, `test`, `rtc`, `chat` |
```
Ruari-Phipps: rtc, test, chat all use local resources (branch id, saving/loading etc)
So I wouldn't put those under builder API

Ruari-Phipps (changes requested):
Yeah following on from what Joey said.
I think we need to allow for some commands to be handed `account_id`, `project_id`, `region`, `branch_id` directly rather than loading it from local file system.
Mainly the functions one here but we could also allow it for others

src/poly/tests/functions_api_test.py
andymohajeri: Function tests now live in the shared cli_test.py and platform_api_test.py

andymohajeri:
@josephcesarperez re: why split this off instead of the standard push/pull workflow — three main reasons:

1. push/pull updates everything on the project, not just the resource being changed. It reads all of flows, entities, topics, agent settings, functions, etc. as one unit. Going through that flow to touch one function pulls the whole project state, which is slow, unnecessary for a single function update, and carries a bigger blast radius — if multiple CI jobs touch different functions, they can clobber each other's unrelated in-flight changes on a shared branch.

2. Some function operations aren't config-sync at all. `execute`, `deployments`, `validate`, `references` are runtime/operational actions — run a function, check deployment history, check orphaned references. There's no push/pull equivalent for any of that.

3. Different consumers want different things. push/pull is for devs working on a project who want git-style workflows (diff, review, branch). Some of the functions endpoints are meant to be called by CI, scripts, and other services — those want a scoped, stateless operation, not a full local checkout.

To be clear, this isn't replacing anything — the existing local `.py`-file/decorator Functions mechanism (`poly/resources/function.py`, synced via push/pull) is untouched and still the right tool for interactive local editing. This is additive, for the automation/headless case.

docs/docs/reference/cli.md
Ruari-Phipps, line 1: The CLI reference has been updated, I would take another look at this please.
Basically this file is for high level, then each sub command has it's own page

src/poly/docs/functions.md
```diff
@@ -4,6 +4,8 @@
 
 Functions are Python files that add deterministic logic to your agent. They can be called by the LLM during conversation, used as flow steps, or run automatically at call start/end.
 
+> To manage functions programmatically over the REST API instead of via local files, see the `poly functions` commands (`poly functions --help`).
+
```
Ruari-Phipps: Is this still true?

src/poly/handlers/interface.py
```diff
@@ -1402,6 +1402,112 @@ def patch_rtc_variables(
...
+
+        Args:
+            region: The region name.
+            project_id: The project ID (agent ID).
+            branch_id: The branch ID.
+
+        Returns:
+            dict: {"valid": bool, "issues": [...]}.
+        """
+        return PlatformAPIHandler.validate_functions(region, project_id, branch_id)
```
Ruari-Phipps, line 1453: What does this do that `poly validate` doesn't?

src/poly/handlers/interface.py
```diff
@@ -1402,6 +1402,112 @@ def patch_rtc_variables(
...
+            project_id: The project ID (agent ID).
+            branch_id: The branch ID.
+            function_id: The function ID.
+
+        Returns:
+            dict: {"references": [...]}.
+        """
+        return PlatformAPIHandler.get_function_references(
+            region, project_id, branch_id, function_id
+        )
```
Ruari-Phipps: Again, might not need a CLI for this. You can just find this by searching {{fn:function_name}}

src/poly/handlers/interface.py
```diff
@@ -1402,6 +1402,112 @@ def patch_rtc_variables(
...
+
+        Args:
+            region: The region name.
+            project_id: The project ID (agent ID).
+            branch_id: The branch ID.
+
+        Returns:
+            dict: {"deployments": [...]}.
+        """
+        return PlatformAPIHandler.list_function_deployments(region, project_id, branch_id)
```
Ruari-Phipps: Not against this completely, but not 100% sure if this is needed. Usually functions are deployed on use? So starting a chat or running tests etc.
This might be something we should just hide from the user

Does executing functions need to deploy first?

src/poly/handlers/interface.py
andymohajeri: poly validate checks local project files/schema entirely offline. poly functions validate is a server-side check. It looks at a branch's deployed functions for orphaned refs and syntax errors.

src/poly/handlers/interface.py
andymohajeri: My gut is to keep this. The returned value shows which flow steps use a function in a branch you haven't pulled locally (a colleague's branch, staging, etc.) via --branch_id.

src/poly/handlers/interface.py
andymohajeri: Confirmed and cutting both: execution already deploys draft functions as a side effect server-side (deploy-on-use). Manual deploy is a branch-level concept (poly branch's existing deploy path), not a per-function one, so there's no function-scoped deploy endpoint to wrap here. Good catch.

src/poly/docs/functions.md
andymohajeri: Corrected it and reworded it to match the scoped-down command.

docs/docs/reference/cli.md
andymohajeri: Finished the migration and removed the duplicate inline sections. Added functions to its own page.

andymohajeri:
Added --region/--project_id/--branch_id passthrough in a prior commit.

docs/docs/reference/cli/functions.md
```diff
@@ -0,0 +1,75 @@
...
+description: Reference for the `poly functions` command.
+---
+
+# `poly functions`
+
+Run and validate Functions using the public Functions REST API, scoped to the project's current branch. `poly functions` requires a subcommand.
+
+!!! note "Local files remain the source of truth"
+
+    Creating, editing and deleting Functions is done via the local `functions/*.py` files synced by `poly pull`/`poly push` (see [Functions](../resources/functions.md)) — that mechanism already covers CRUD, including flow-scoped transition functions, function steps and latency control, and keeps changes reviewable in a branch diff before they're pushed. `poly functions` is additive: it covers what push/pull can't — running a function and validating it — via direct REST calls that don't touch local files.
```
Ruari-Phipps: Maybe reword? Is this what should be in a docs page?

docs/docs/reference/cli/functions.md
```diff
@@ -0,0 +1,75 @@
...
+
+Examples:
+
+~~~bash
+poly functions execute <function_name_or_id> --args '{"x": 1}'
```
Ruari-Phipps: Can we just have name

docs/docs/reference/cli/functions.md
```diff
@@ -0,0 +1,75 @@
...
+
+## `poly functions execute`
+
+Execute a Function and print its return value, logs and runtime.
+
+Examples:
+
+~~~bash
+poly functions execute <function_name_or_id>
+poly functions execute <function_name_or_id> --args '{"x": 1}'
```
Ruari-Phipps: Maybe we should have a not how this works with the conv object?

Ruari-Phipps: Also is this global functions or does this also include flow functions?

src/poly/cli_commands/functions.py
```diff
@@ -0,0 +1,268 @@
...
+        for candidate in functions:
+            if function in (candidate.get("id"), candidate.get("name")):
+                return candidate["id"]
+
+        msg = f"No function named or with ID '{function}' found on this branch."
+        if output_json:
+            json_print({"success": False, "error": msg})
+        else:
+            error(msg)
+        sys.exit(1)
```
Ruari-Phipps, line 188: Ah yeah if we error on missing function we should update the docs

Ruari-Phipps (commented):
Implementation looks good. Just some notes on the docs

docs/docs/reference/cli/functions.md
andymohajeri: Reworded to two direct sentences, dropped the PR-rationale framing.

docs/docs/reference/cli/functions.md
andymohajeri: Done — examples now use <function_name> everywhere; the argument table still notes ID also works.

docs/docs/reference/cli/functions.md
andymohajeri: Added a note on how conv actually behaves (fresh per-run, no live caller/state) plus a doesn't-work-this-way example, and clarified this only covers global functions, not flow-scoped ones.

src/poly/cli_commands/functions.py
andymohajeri: Added an info admonition documenting the not-found error.

docs/docs/reference/cli.md
andymohajeri: Confirmed done. Resolving.

src/poly/docs/functions.md
andymohajeri: Confirmed accurate after the scope cut. Resolving.

src/poly/handlers/interface.py
andymohajeri: Confirmed accurate. Resolving.

src/poly/handlers/interface.py
andymohajeri: Update: after further discussion this was actually cut entirely, not kept — function_id/name isn't surfaced usefully enough elsewhere to justify it. Resolving.

src/poly/handlers/interface.py
andymohajeri: Confirmed cut. Resolving.

Ruari-Phipps (approved):

PR #247, opened 2026-07-30
feat: branch sync, history, rename, tag/untag, and archive/restore

Ruari-Phipps:
## Summary

Adds six `poly branch` subcommands — `sync`, `history`, `rename`, `restore`, `tag`, `untag` — plus `branch list --archived`, and teaches `branch create`/`merge`/`current`/`list` about branch lineage and the project's deployment mode. Commands the platform only exposes under simplified deployments are gated behind the `deployment_simplification` feature flag. Also makes `poly deployments list`/`show`/`promote`/`rollback` aware of that same deployment mode.

## Motivation

The platform's simplified deployment model introduces branch lineage (branches can be created from other branches), staging tags, and soft-archive/restore with a 30-day window. The ADK CLI had no support for any of it, and `poly branch merge` assumed every branch's parent was `main`. Once a project moves to simplified deployments, its old `sandbox` deployment history is also frozen in place — merges to main go straight to `live` — so the existing `deployments` commands needed to stop defaulting to `sandbox`.

## Changes

**New commands**

- `poly branch sync` — merge the parent branch's changes into the current branch, reusing the existing merge-conflict UX (`-i`/`--interactive`, `--resolutions`)
- `poly branch history [--branch-name] [--limit]` — merge history for a branch, 10 entries by default
- `poly branch rename [new_name]` — rename the current branch
- `poly branch restore [branch]` — restore a soft-deleted branch from the archive
- `poly branch tag` / `poly branch untag` — tag the current branch to deploy it to staging, or remove the tag
- `poly branch list --archived` — list soft-deleted branches instead of active ones

**Changed behaviour**

- `branch create` now respects the project's deployment mode: one active branch in `simple`, main-only in `releases`, and branching off a direct child of main (max depth 2) in `releases_branches`. This replaces the previous unconditional "branches can only be created from main" guard. A new `--from <branch>` flag lets the source branch be named explicitly instead of always defaulting to whatever branch is currently checked out; the deployment-mode guards validate against the named source, not the current branch.
- `branch merge` merges into the branch's actual parent instead of always `main`, switches to that parent afterwards, and shows a confirmation plus "now live" messaging when merging into main under simplified deployments. Merges into a branch other than `main` no longer require a merge message.
- `branch current` also prints the parent branch, suppressed when the parent is `main` to avoid noise in the common case. `--json` always includes `parent_branch`.
- `branch list` and the `switch`/`delete` interactive pickers render lineage as an indented tree in `releases_branches` mode, and show staging tags.
- `branch` subcommands are regrouped into lifecycle and inspection groups in `--help`, and stale help text/epilogs are corrected.
- `deployments list`/`show` default to the `live` environment instead of `sandbox` for projects using simplified deployments (`--env` still overrides). `show` no longer prints a "No intermediate deployments" placeholder when the included list is empty, since post-migration deployments legitimately have none.
- `deployments promote` is refused outright for projects using simplified deployments — merging to main deploys straight to live, so there is no sandbox → pre-release → live ladder left to promote along, and the platform does not reject the call itself.
- `deployments rollback` targets `live` rather than `sandbox` for projects using simplified deployments, since that's the only environment the platform accepts a rollback target for.

**Feature-flag gating**

- New `PosthogHandler` and `AgentStudioInterface.feature_flag_enabled`, backing `AgentStudioProject.using_simplified_deployments`.
- `using_simplified_deployments` now also checks convergence: even with the feature flag on, it only reports `True` once the project's `live` deployment head is at least as recent as its `sandbox` head (i.e. `live` has caught up and sandbox is no longer receiving new deployments). This avoids treating a project as simplified mid-migration, while it still has newer sandbox deployments that haven't reached live.
- New `require_deployment_simplification` helper gates `sync`, `tag`, and `untag` — the three endpoints the platform 404s when the flag is off. Previously these surfaced a bare `API error: 404 Client Error`; they now exit 1 with an explanatory message in both human and `--json` modes.
- New `DeploymentMode` enum and `AgentStudioProject.deployment_mode`, read from the project's `config.deployment_mode` and defaulting to `releases`.

**Supporting API surface**

`sync_branch`, `get_branch_history`, `rename_branch`, `list_archived_branches`, `restore_branch`, `tag_branch`, `untag_branch`, `get_project`, and `create_branch(..., source_branch_id)` across the SDK, sync client, and interface. New console renderers for branch trees, archived branches, and merge history.

**Other**

- Adds `posthog==7.35.4`.
- README `poly branch` section and `poly branch --help` examples updated to cover the new commands, including `--from`.

## Follow-ups (tracked separately, not silently dropped)

- The `sandbox` environment does not exist under simplified deployments, but it is still the hardcoded default for `diff`, `chat`, and the `rtc` commands, so those return empty results rather than an explanation. Fixing that also means introducing a shared environment constant — the three-environment literal is currently duplicated at 15+ sites — and renaming `sandbox`/`pre-release` to the branches/staging vocabulary the product now uses.

## Test strategy

## Checklist

- [x] `pytest` passes (1208 passed, 61 subtests)

### DIFF
--- a/src/poly/cli_commands/branch.py
+++ b/src/poly/cli_commands/branch.py
@@ -663,26 +663,44 @@ class BranchCommand(BaseCommand):
 
     @classmethod
     def get_current_branch(cls, base_path: str, output_json: bool = False) -> None:
-        """Get the current branch of the Agent Studio project."""
+        """Get the current branch of the Agent Studio project, and its parent if it has one."""
         from poly.output.console import plain, warning
 
         project = load_project(base_path, output_json=output_json)
 
-        current_branch = project.get_current_branch()
-        if output_json:
-            json_output = {
-                "current_branch": current_branch,
-            }
-            json_print(json_output)
-            return
+        current_branch, branches = project.get_branches()
 
         if current_branch is None:
-            warning(
-                "Current local branch does not exist in Agent Studio. "
-                "It may have been deleted or merged."
-            )
+            if output_json:
+                json_print({"current_branch": None, "parent_branch": None})
+            else:
+                warning(
+                    "Current local branch does not exist in Agent Studio. "
+                    "It may have been deleted or merged."
+                )
             return
+
+        parent_branch_id = branches.get(current_branch, {}).get("parentBranchId")
+        parent_branch = (
+            next(
+                (
+                    name
+                    for name, meta in branches.items()
+                    if meta.get("branchId") == parent_branch_id
+                ),
+                parent_branch_id,
+            )
+            if parent_branch_id
+            else None
+        )
+
+        if output_json:
+            json_print({"current_branch": current_branch, "parent_branch": parent_branch})
+            return
+
         plain(f"Current branch: [bold]{current_branch}[/bold]")
+        if parent_branch:
+            plain(f"Parent branch: [bold]{parent_branch}[/bold]")
 
     @classmethod
     def branch_delete(
--- a/src/poly/cli_commands/branch.py
+++ b/src/poly/cli_commands/branch.py
@@ -1387,6 +1387,18 @@ class BranchCommand(BaseCommand):
 
         project = load_project(base_path, output_json=output_json)
 
+        if not project.using_simplified_deployments:
+            if output_json:
+                json_print(
+                    {
+                        "success": False,
+                        "error": "Branch sync is only available for projects using simplified deployments.",
+                    }
+                )
+            else:
+                error("Branch sync is only available for projects using simplified deployments.")
+            sys.exit(1)
+
         branch_name = project.get_current_branch()
         ctx = console.status("[info]Syncing branch...[/info]") if not output_json else nullcontext()
         with ctx:
--- a/src/poly/cli_commands/branch.py
+++ b/src/poly/cli_commands/branch.py
@@ -370,57 +370,6 @@ class BranchCommand(BaseCommand):
         ).completer = cls._branch_name_completer
         branch_status_parser.set_defaults(branch_subcommand="status")
 
-        # -- history ──
-        branch_history_parser = branch_subparsers.add_parser(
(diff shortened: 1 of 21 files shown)


Wiz1991 (approved):

src/poly/handlers/posthog.py
```diff
@@ -0,0 +1,99 @@
+"""Client for the PolyAI Posthog tenant, used for feature flags in the PolyAI ADK CLI.
+
+Copyright PolyAI Limited
+"""
+
+import logging
+from typing import TYPE_CHECKING
+
+POSTHOG_HOST = "https://eu.i.posthog.com"
+POSTHOG_KEY = "phc_kS54QZyZRqi9T77rWEUfJ49vVYY4ADKEPnRUrJ7RNnZ6"
```
jamesosullivan-polyai, line 10: nit: can we put this in an env var?

src/poly/cli_commands/branch.py
```diff
@@ -1140,3 +1424,397 @@ def branch_status(
...
+        branch_id = branches.get(branch_name, {}).get("branchId")
+        if not branch_id:
+            if output_json:
+                json_print({"success": False, "error": f"Branch '{branch_name}' does not exist."})
+            else:
+                warning(f"Branch '{branch_name}' does not exist.")
+            return
+
+        history = project.get_branch_history(branch_id)
+        history = history[:limit]
```
jamesosullivan-polyai, line 1610: nit: this still fetches everything from the backend. Perhaps in the future we'd want to paginate the backend and pass the limit

jamesosullivan-polyai (approved):
lgtm!

src/poly/handlers/posthog.py
Ruari-Phipps: This is the public client key. We need to have it exposed here

PR #256, opened 2026-08-05
feat: add SIP header overrides to chat

polyai-moaed:
## Summary

Adds a repeatable `poly chat --sip-header NAME=VALUE` option that simulates SIP headers on chat conversations, so agent behaviour that reads `conv.sip_headers` can be exercised without arranging a real SIP call.

## Motivation

Testing header-dependent agent logic today requires a real SIP call. This gives developers a quick local way to inject agent-visible header values from the chat CLI. Note the option is additive and simulates agent-visible values only; it does not reproduce SIP transport or carrier behaviour, and header injection depends on server-side chat API support rolling out separately.

## Changes

- Add a repeatable `--sip-header NAME=VALUE` option to the modular chat command
- Validate header syntax while preserving values that contain `=`
- Forward simulated headers through standard and draft conversation creation
- Document the option

## Test strategy

- [x] Added/updated unit tests (CLI parsing, session forwarding, API payloads)

## Checklist

- [x] `pytest` passes

## Screenshots / Logs

`pytest -q` result: 994 passed, 30 subtests passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,8 +201,13 @@ Start an interactive chat session with your agent:
 poly chat
 poly chat --environment live
 poly chat --channel webchat
+poly chat --sip-header X-Customer-ID=12345 --sip-header X-Language=en-GB
 poly chat --metadata   # show functions, flows, and state each turn
 ```
+
+`--sip-header NAME=VALUE` is repeatable and simulates headers exposed to project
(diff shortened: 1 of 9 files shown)


Ruari-Phipps (approved):
Nice LGTM!

PR #259, opened 2026-08-07
feat: add poly metrics commands (list, export, add, edit, import)

bill-parker:
## Summary

Add `poly metrics` CLI commands for managing custom metrics in Agent Studio projects:
* list
* export / import (YAML)
* add
* edit

## Motivation

Custom metrics are currently only manageable through the Agent Studio UI. This adds full CLI support so teams can script metric setup, export/import metric definitions across projects, and integrate metric management into CI pipelines.

### Commands

- **`poly metrics list`** — display all custom metrics in a Rich table (Name, Type, Active, API, Description) with active/inactive count summary; `--json` for machine-readable output
- **`poly metrics export [file]`** — export all metrics as YAML to stdout or a file; `--json` outputs JSON instead
- **`poly metrics add`** — create a new metric with `--name`, `--type`, `--description`, `--api`, `--expected-values`; omit required flags for interactive mode (name text input, type selector, description prompt, API confirm via `questionary`)
- **`poly metrics edit <name>`** — update an existing metric's `--description`, `--api` (bool), `--active` (bool), or `--expected-values`; `--api`/`--active` accept `true`/`false` or can be used bare to set `true`. When called with just a metric name and no flags, enters **interactive mode**: fetches the metric, displays current values, and prompts with `questionary.checkbox()` to select which fields to update. `--expected-values` is only offered for string-type metrics. Friendly error when metric not found (404).
- **`poly metrics import <file>`** — bulk-import metrics from a YAML file; creates new metrics and skips existing ones; `--dry-run` previews what would be created/skipped and warns about remote-only metrics that won't be deleted

### API flag workaround

The server's `create_custom_metric` route hardcodes `api=False`, ignoring the client-supplied value (the AS UI also doesn't set this on create). When `--api` is passed, under the hood the the CLI follows up with a PATCH to set `api=True`.

### Implementation notes

- Platform API: 5 new methods on `PlatformAPIHandler`; `make_request` extended with `files` (multipart upload) and `response_format` (YAML parsing) params . These were needed for import and export YAML as file handling only previously supported JSON.
- Tests: 30 unit tests across 8 test classes

import/export flow was chosen here as the implementation details of metrics require that metrics
have project scope (not branch scope) and cannot be deleted once written.

In this case, using a local file to manage additions and changes could give way to merge issues
and user confusion where deleted items are not actually removed.
For this reason I have opted to model this feature on the platform features,
with one exception that API can be set on metric creation.

## Test strategy

- [x] Manual CLI testing (`poly metrics` for all subcommands)

**Manual test spreadsheet**: [Google Sheet](https://docs.google.com/spreadsheets/d/1ahhwLpTKTt_pApH7Z_oESQkX8ki-vj2GrHf2ABO75Zw/edit?gid=246662677#gid=246662677)

## Checklist

- [x] `pytest` passes

### `poly metrics list`
Screenshot 2026-07-31 at 15 06 03

### `poly metrics add` (interactive)
Screenshot 2026-08-04 at 16 53 09

### `poly metrics add` — API flag prompt
Screenshot 2026-07-31 at 15 17 40

### `poly metrics add --expected-values` (JSON output)
Screenshot 2026-07-31 at 15 21 57

### `poly metrics add --api` (JSON output)
Screenshot 2026-07-31 at 15 30 58


src/poly/handlers/interface.py
Ruari-Phipps, line 1: My main comment is on general structure.
I feel like we should be running this through the `project.py`
So cli -> project -> interface -> api
Adds a little more but thats the general pattern I've been aiming for so far. I might adjust this going forward.

Also, would be good to have some of the logic in here being moved into the project.py. This file is basically just a forwarder to the correct endpoint and some error handling. Stuff like validating and reading the files should be in there

src/poly/tests/metrics_test.py
Ruari-Phipps, line 1: Should this be in cli_test?

northdpole:
👋 The failing `trufflehog` check on this PR is a **false positive — there is no credential here, and nothing needs rotating.**

The scanner flagged a test function name, [`src/poly/tests/metrics_test.py:321`](https://github.com/polyai/adk/blob/main/src/poly/tests/metrics_test.py#L321):

```python
def test_add_duplicate_metric_friendly_error(self, mock_load, mock_create, mock_error):
```

TruffleHog's Lob detector matches `\b((live|test)_[a-zA-Z0-9_]{35})\b` with no keyword-proximity requirement, and its keyword prefilter is literally `{"live_", "test_"}` — the pytest naming convention. So any `test_*` identifier that is exactly 40 characters long matches. This one is. It gets reported as *verified* because the detector counts HTTP 403/422 from `api.lob.com` as success, so a function name passes verification.

Nothing to change on your side — the name is fine, please don't rename it.

#263 fixes this at the source (excludes the Lob detector, which we don't use, and makes the scanner actually print the file, line and match instead of a bare `Found verified Lob result`). Once that merges, rebase on `main` and this check will go green.

Sorry for the noise — the output gave you no way to tell a real leak from this.

src/poly/tests/metrics_test.py
bill-parker: have moved tests into relevant layers!

src/poly/handlers/interface.py
bill-parker: Have moved most read/write logic into project, leaving interface as thin a layer as possible

src/poly/cli_commands/metrics.py
```diff
@@ -0,0 +1,582 @@
...
+                output_json=args.json,
+            )
+
+    @classmethod
+    def metrics_list(cls, base_path: str, output_json: bool = False) -> None:
+        """List all custom metrics for the project."""
+        project = load_project(base_path, output_json=output_json)
+        metrics = AgentStudioInterface.get_custom_metrics(
+            project.region, project.account_id, project.project_id
+        )
```
Ruari-Phipps: Nit. Have this be a project function too

src/poly/output/console.py
```diff
@@ -183,6 +183,46 @@ def print_agents(agents: list[dict[str, Any]]) -> None:
     console.print(table)
 
 
+def print_metrics(metrics: list[dict[str, Any]]) -> None:
```
Ruari-Phipps, line 186: I added a pager system for some other print functions, would be good here too so you can scroll through them

src/poly/cli_commands/metrics.py
```diff
@@ -0,0 +1,582 @@
...
+            result = project.create_custom_metric(data)
+        except ValueError as e:
+            if output_json:
+                json_print({"success": False, "error": str(e)})
+            else:
+                error(str(e))
+            sys.exit(1)
+        except requests.HTTPError as e:
+            if e.response is not None and e.response.status_code == 409:
+                msg = f"Metric '{name}' already exists."
```
Ruari-Phipps: Might be good to have some of this handling within the project layer?

src/poly/cli_commands/metrics.py
```diff
@@ -0,0 +1,582 @@
...
+            result = project.update_custom_metric(name, data)
+        except ValueError as e:
+            if output_json:
+                json_print({"success": False, "error": str(e)})
+            else:
+                error(str(e))
+            sys.exit(1)
+        except requests.HTTPError as e:
+            if e.response is not None and e.response.status_code == 404:
+                msg = f"Metric '{name}' not found."
```
Ruari-Phipps: Same here

src/poly/cli_commands/metrics.py
```diff
@@ -0,0 +1,582 @@
...
+
+        Returns:
+            A dict of fields to update.
+        """
+        import questionary
+
+        metrics = AgentStudioInterface.get_custom_metrics(
+            project.region,
+            project.account_id,
+            project.project_id,  # type: ignore[attr-defined]
```
Ruari-Phipps: Why type thing?

Ruari-Phipps (approved):
LGTM!

PR #267, opened 2026-08-12
feat(test-suite): support mock call context on test cases

Wiz1991:
## Summary

Adds caller number, SIP headers and integration attributes to the test case resource so `poly pull` and `poly push` round-trip them, with docs in `src/poly/docs/tests.md`. Until now the ADK silently dropped all three.

**Stacked on #266** — its base should be retargeted to `main` once that merges.

## Motivation

Agent Studio can set mock call context on a test case; the ADK could not see or send it. Pulling a project omitted the fields from the YAML and pushing left them untouched on the server.

Closes [DEVP-580](https://linear.app/poly-ai/issue/DEVP-580)

## Changes

- `caller_number` on the `TestCase` resource — a scalar on `Create_TestCase` / `Update_TestCase`. Always sent, empty string included, since proto3 has no null and clearing it otherwise would not stick
- `TestCaseSipHeaders` and `TestCaseIntegrationAttributes` subresources, each with its own `set_` command, on the existing `TestCaseTags` pattern
- Both subresources are always constructed, even when empty — `get_new_updated_deleted_subresources` compares subresources, so an absent one cannot differ from the populated one it replaces and clearing a value would never push
- `from_projection` reads the camelCase projection keys (`callerNumber`, `sipHeaders`, `integrationAttributes`)
- Type handling so the YAML path matches the platform (see below)
- `src/poly/docs/tests.md` — a "Mock call context" section with a worked example, per-field notes and the quoting rules

## Test strategy

- [x] Added/updated unit tests — 21 new
- [ ] N/A

**1029 passed**, ruff clean. Coverage of `test_suite.py` +3.4%.

Four behaviours are mutation-tested — each fails when the code it covers is reverted:

| Test | Reverting |
|---|---|
| `test_struct_doubles_are_written_back_as_ints` | the float normalisation |
| `test_clearing_subresources_still_pushes` | guarding the comparison on non-emptiness |
| `test_sip_header_bools_use_wire_casing` | `_header_value` back to `str()` |
| `test_validate_rejects_dates_with_an_actionable_message` | the date branch |

## Checklist

- [x] `pytest` passes
- [x] No breaking changes to the `poly` CLI interface — additive fields only

## Screenshots / Logs

YAML types unquoted scalars, so it can produce values the Studio UI's text inputs never can. I probed every type against the generated protos rather than assuming; `string`, `number`, `float`, `bool`, `null`, `list` and nested `map` all round-trip correctly. Four cases did not:

| YAML | Was | Now |
|---|---|---|
| `x-flag: true` (SIP header) | `"True"` — Python's casing | `"true"` |
| `x-none:` (SIP header) | `"None"` | `""` |
| `expiry: 2026-08-12` (attribute) | `ValueError: Unexpected type` | rejected in `validate()` with the path and the quoted form |
| `caller_number: +447700900000` | `TypeError: bad argument type` | rejected — see below |

A `Struct` also stores every number as a double, so an attribute pushed as `2` returns as `2.0` and would write `2.0` back to YAML, giving a spurious diff on every pull after a push. Integral floats are folded back to `int`; genuine decimals are untouched.

The caller number one is the subtle one. YAML reads `+447700900000` as the int `447700900000` — **dropping the `+`** — and that is indistinguishable from someone writing no `+` at all:

```
caller_number: +447700900000   -> 447700900000   str() -> '447700900000'
caller_number: 447700900000    -> 447700900000   str() -> '447700900000'
```

So `str()` would silently send a different number. It is rejected with a message saying to quote it. Caller number is also now trimmed on push, matching the platform entity, where a trailing space changes the agent-memory identifier.

## Still missing after this

`severity`, `simulated_at` and `api_mocks` remain unsupported by the resource layer, despite the latter two now being in the protos. Those are the same six fields [BLD-811](https://linear.app/poly-ai/issue/BLD-811) reports missing from the environment comparison renderer — every consumer of the TestCase model except the main form has fallen behind by the same set. Worth a follow-up.


Wiz1991:
Pushed `fa8951a` after checking how each type actually serialises, rather than assuming the YAML path matched the UI's. It didn't, in two places — both caused by YAML typing unquoted scalars where the UI only ever has text inputs.

**SIP headers sent Python's bool casing.** `x-flag: true` parses to a Python `bool`, and `str()` made that `"True"` — which neither the UI nor a carrier ever produces. Now `"true"`/`"false"`, with `None` as empty rather than the string `"None"`, and dates as ISO text.

| YAML | was sent | now |
|---|---|---|
| `x-flag: true` | `"True"` | `"true"` |
| `x-none:` | `"None"` | `""` |
| `x-date: 2026-08-12` | `"2026-08-12"` | unchanged |

**Integration attributes crashed on YAML-only types.** A `Struct` carries exactly JSON's types, so an unquoted `expiry: 2026-08-12` — a `datetime.date` — reached `Struct.update` and raised:

```
ValueError: Unexpected type
```

No key, no file, no suggestion. A non-string key was worse: `TypeError: bad argument type for built-in operation`. Both are now caught in `validate()` with the full path and, for dates, the quoted form to use:

```
Integration attribute 'integration_attributes.booking.slots[0].at' is a date,
which the agent cannot receive — quote it to send it as text: '2026-08-12'
```

I probed every type against the generated protos rather than reasoning about it. `string`, `number`, `float`, `bool`, `null`, `list` and nested `map` all round-trip correctly; `date`, `datetime` and non-string keys were the only rejects, and they are now rejected in the right place with the right message.

6 further tests (19 in this area, 1027 overall). Bool casing and the date message are both mutation-tested — reverting `_header_value` to `str()` fails the first, and dropping the date branch fails the second, since the generic fallback still rejects it but loses the actionable suggestion.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Ruari-Phipps (approved):
Nice! LGTM!

PR #284, opened 2026-08-25
feat: support API integration mocks on test cases

CollwynB:
## Summary

Adds `api_mocks` support to test cases — mock an API integration operation's response (status/body/headers, with a `repeat` sequence) so flows that branch on API responses can be tested deterministically during simulation.

## Motivation

Closes AM-960

## Changes

- New `ApiResponse`, `ApiResponseRule`, `TestCaseApiMocks` dataclasses in `test_suite.py`, wired into `TestCase` YAML load/dump, `from_projection`, and `push` diffing, following the existing `SubResource` pattern (`TestCaseIntegrationAttributes`, `TestCaseSipHeaders`).
- Push-time validation: integration name must match a known `api_integration` resource, `status` must be 100–599, `body` follows the same type rules as `integration_attributes`, each operation needs at least one response rule, and `repeat` must be a positive integer or `-1` (respond forever) — `-1` only valid on the last rule in an operation's list.
- Cross-checked `repeat`/`-1` semantics and rename/delete cascading behavior against the platform runtime (`genai_lambda_runtime`) and the web app (`platform_ui/apps/agent-stream` + `jupiter`) to make sure the CLI's validation and docs match actual platform behavior.
- Docs: new "API mocks" section in `tests.md` with a YAML example, cross-linked from `api_integrations.md`.

## Test strategy

## Checklist

- [x] `pytest` passes (1332 passed)
- [x] No breaking changes to the `poly` CLI interface
- [x] Commit messages follow conventional commits

## Screenshots / Logs

N/A


src/poly/handlers/protobuf/__init__.py
Ruari-Phipps, line 1: Re-instate

src/poly/handlers/protobuf/poly_platform/ragdoll/pb/BUILD.bazel
Ruari-Phipps, line 1: Delete

src/poly/resources/test_suite.py
```diff
@@ -703,6 +906,60 @@ def validate(self, resource_mappings: list[ResourceMapping] = None, **kwargs):
...
+                            or rule.repeat < -1
+                        ):
+                            raise ValueError(
+                                f"{label}: repeat '{rule.repeat}' must be a positive integer or -1."
+                            )
+                        if rule.repeat == -1 and index != len(rules) - 1:
+                            raise ValueError(
+                                f"{label}: repeat=-1 (respond forever) is only valid on the "
+                                "last response rule for an operation."
+                            )
```
Ruari-Phipps: I would move this into API mocks own validate function. Just to clean it up

src/poly/resources/test_suite.py
```diff
@@ -754,8 +1056,11 @@ def get_new_updated_deleted_subresources(
                 updated.append(self.sip_headers)
             if old_resource.integration_attributes != self.integration_attributes:
                 updated.append(self.integration_attributes)
+            mock_updates, mock_deletes = self._diff_api_mocks(old_resource.api_mocks.mocks)
+
+        updated.extend(mock_updates)
 
-        return [], updated, []
+        return [], updated, mock_deletes
```
Ruari-Phipps: Have a general deletes list, to make it easily extendable

src/poly/docs/api_integrations.md
Ruari-Phipps, line 1: Can we also update the other docs files too (outside src/poly)
There might also be a general testing page to update as well

src/poly/tests/resources_test.py
Ruari-Phipps, line 1: Can we also add an example to the `test_project` too 👍

Ruari-Phipps (commented):
Generate changes look good I think, some clean up and adding to the project tests would be good!

docs/docs/development/testing.md
```diff
@@ -46,6 +46,8 @@ For the test-case format, the available assertions, and worked examples, see the
 
     Interactive chat is good at the common path and bad at the awkward ones — an unavailable API, a caller who changes their mind, a variant that only applies to one site. Those are worth encoding as test cases precisely because reaching them manually is tedious.
 
+    For the "unavailable API" case specifically, `api_mocks` lets a test case force a named [API integration](../reference/resources/api_integrations.md) operation to return a given response, so a flow that branches on it runs deterministically instead of depending on a real (and possibly flaky) external call. See [API mocks](../reference/resources/tests.md#api-mocks).
+
```
Wiz1991, line 50: what is this abut?

docs/docs/reference/resources/api_integrations.md
```diff
@@ -149,4 +149,11 @@ response = conv.api.salesforce.update_contact(
     Call API integrations from advanced-step and transition-function prompts.
     [Open flows](./flows.md)
 
+-   **Tests**
+
+    ---
+
+    Mock an operation's response with `api_mocks` so a flow that branches on it can be tested deterministically.
+    [Open tests](./tests.md)
```
Wiz1991: why does teh flow matter here?

src/poly/docs/api_integrations.md
```diff
@@ -89,4 +89,8 @@ response = conv.api.salesforce.update_contact(
     params={"id": "123"},
     json={"phone_number": "456"}
 )
-```
\ No newline at end of file
+```
+
+## Testing with mocked responses
+
+A flow that branches on an operation's response (e.g. slot negotiation, availability checks) needs that response to be deterministic to test reliably. Test cases can mock an operation's response per test — see [Tests: API mocks](tests.md#api-mocks).
```
Wiz1991: api integrations are not only for flows tho?

Wiz1991 (changes requested):

docs/docs/development/testing.md
CollwynB: it just follows on from this
`Interactive chat is good at the common path and bad at the awkward ones — an unavailable API...`

Unavailable API in this case, I guess is referring to any reason you might not be able to use an API in a given test.

docs/docs/reference/resources/api_integrations.md
CollwynB: Flow in this case doesn't mean a flow tool as configured in the platform, but the actual flow of conversation

src/poly/docs/api_integrations.md
CollwynB: Same as above

CollwynB: maybe I should just rename flow to conversation in this contexts?

docs/docs/reference/resources/api_integrations.md
Wiz1991: hm, is this common across docs?

src/poly/docs/api_integrations.md
Wiz1991: i find it a bit confusing, it overlaps with actual flows

CollwynB: I agree, but this seems to be the wording in other places already, `tests.md` for example:
>  Unlike SIP headers, these **keep their type** through to the agent, so a flow branching on `retry_count > 2` sees a number rather than the text `"2"`. Text, numbers, `true`/`false`, `null`, lists and nested maps are all supported.

> - Set mock call context only where the flow reads it. A test that carries headers no node inspects is noise.
> - Keep one test focused on one behaviour or flow path.

Wiz1991 (approved):

docs/docs/reference/resources/api_integrations.md
Ruari-Phipps: ?
```suggestion
    Mock an operation's response with `api_mocks` so a conversation that branches on it can be tested deterministically.
    [Open tests](./tests.md)
```

src/poly/handlers/protobuf/BUILD.bazel
Ruari-Phipps, line 1: delete

src/poly/resources/test_suite.py
```diff
@@ -356,6 +365,227 @@ def build_delete_proto(self) -> None:
...
+    def from_dict(cls, data: dict[str, dict[str, list[dict]]] | None) -> "TestCaseApiMocks":
+        data = data or {}
+        mocks = {
+            integration_name: {
+                operation_name: [ApiResponseRule.from_dict(rule) for rule in rules or []]
+                for operation_name, rules in (operations or {}).items()
+            }
+            for integration_name, operations in data.items()
+        }
+        return cls(mocks=mocks)
```
Ruari-Phipps, line 525: Maybe should do some type checking here to ensure data is in the correct shape?

src/poly/resources/test_suite.py
```diff
@@ -490,6 +740,12 @@ def __init__(
             self.integration_attributes = TestCaseIntegrationAttributes(
                 resource_id=resource_id, name="integration_attributes"
             )
+        if isinstance(api_mocks, TestCaseApiMocks):
+            self.api_mocks = api_mocks
+        elif api_mocks:
+            self.api_mocks = TestCaseApiMocks.from_dict(api_mocks)
```
Ruari-Phipps: When loading back from disk, this will fail as it will be loading back the whole object rather than the just the mocks.
See SIPHeaders above

Ruari-Phipps (approved):

PR #288, opened 2026-08-26
feat: handle auth-filtered projections without dropping or stranding local files

Ruari-Phipps:
## Summary

The API filters projections per permission slice: a user without read access to a slice
gets a skeleton carrying only identity fields, with the substantive fields stripped. This
makes every resource type tolerate that shape, and keeps withheld resources as lightweight
mappings rather than half-built resources, so references to them still resolve to a name.

## Motivation

Pulling as a restricted reader either crashed on a missing field or wrote out resources
built from a skeleton, corrupting them on the next push. Withheld resources also have to
survive in the status file: without them, every `{{entity:...}}`-style reference to a
resource the user cannot read renders as a raw id, and the file then looks locally
modified on every command.

## Changes

- Every `from_projection` now detects an auth-filtered slice and either drops the resource
  or keeps it as an identity-only stub, instead of assuming fields are present.
- Withheld resources are separated out of the resource map into `slim_resources`, a list of
  `ResourceMapping`. They resolve references but are never iterated for a file, saved, or
  pushed. `load_resources_from_projection` and `pull_resources` return them alongside the
  resource map.
- `ResourceMapping` gained `to_dict`/`from_dict`, storing `resource_type` by its registered
  name so slim mappings persist in the status file. An unregistered type is dropped on read
  rather than raising.
- Slim resources are excluded from `file_structure_info` — they have no file on disk, so a
  baseline entry would make `find_new_kept_deleted` report them deleted on every run.
- `pull_resources`, `pull_deployment_resources`, `pull_branch_resources`,
  `get_template_resources`, `get_remote_resources_by_name` and `_resolve_branch_fork_point`
  all return their slim mappings alongside the resource map, and every caller was updated
  to match.
- Fixed multi-resource YAML deletions never reaching disk when an entire type is absent from
  the incoming projection. The deletions were batched into the file cache, which was then
  discarded when the pull returned. The leftover cache entry was also stamped with the file's
  pre-write mtime, so the staleness check treated it as fresh and every later read in the
  process saw a file state that was not on disk. Withheld types take this path, so a force
  pull had to be run twice before `poly diff` came back clean.

## Test strategy

## Checklist

- [x] `pytest` passes (1372 passed, 159 subtests)


Ruari-Phipps:
Moved into several PRs 
298, 299, 300

PR #291, opened 2026-08-27
fix: sort resource collections in YAML and register fetch command

Ruari-Phipps:
## Summary

Sorts collections alphabetically during YAML serialization so pulls produce stable, diff-friendly output. Also registers `FetchCommand` in the CLI command list — it was implemented but never wired up, so `poly fetch` was unavailable.

## Motivation

Several collections are built by iterating a map in the platform projection — flow step conditions, API integration operations, test case function call assertions and their arguments, and the integration/operation keys of test case API mocks. Their serialized order followed whatever the map iteration gave, so unrelated pulls could reorder blocks in a resource's YAML and create noisy diffs.

Separately, the `fetch` command existed in `cli_commands/sync.py` but was missing from `COMMANDS`, so it never appeared in the CLI.

## Changes

- `FlowStep.to_yaml_dict` sorts conditions by `name` (matches the existing `sorted(self.extracted_entities)` behaviour)
- `ApiIntegration.to_yaml_dict` sorts operations by `name`
- `FunctionCallAssertion.to_yaml_dict` sorts arguments by `parameter_name`, and `TestCaseAssertion.to_yaml_dict` sorts function call assertions by `name`
- `TestCaseApiMocks.to_yaml_dict` sorts integration and operation names; the rules within an operation keep their order, since they are a sequence and `repeat` depends on it
- Add `FetchCommand` to the `COMMANDS` list in `cli.py` so `poly fetch` is registered
- Add unit tests covering each ordering, including one asserting mock rule order is preserved
- `uv.lock` version bump picked up from the 0.44.4 release

## Test strategy

- [x] Manual CLI testing (`poly --help` now lists `fetch`)

## Checklist

- [x] `pytest` passes (1356 passed)

## Screenshots / Logs

```
$ uv run poly --help | grep fetch
    fetch               Fetch the latest project state from Agent Studio
```

### DIFF
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -24,6 +24,7 @@ from poly.cli_commands.review import ReviewCommand
 from poly.cli_commands.rtc import RTCCommand
 from poly.cli_commands.sync import (
     DiffCommand,
+    FetchCommand,
     FormatCommand,
     PullCommand,
     PushCommand,
@@ -45,6 +46,7 @@ COMMANDS = [
     StudioCommand,
     ProjectCommand,
     TemplateCommand,
+    FetchCommand,
     PullCommand,
     PushCommand,
     StatusCommand,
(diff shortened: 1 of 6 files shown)


oeisenberg (approved):
See it, say it, sorted

PR #294, opened 2026-08-27
feat: add persona to agent settings (DEVP-624)

toswe:
## Summary

Replaces the separate Personality and Role settings in ADK with the single free-text `persona` resource that superseded them in Agent Studio. `agent_settings/persona.txt` now pulls, pushes and diffs; `personality.yaml` and `role.yaml` are gone, and a migration deletes them from existing project directories.

## Motivation

[DEVP-624](https://linear.app/poly-ai/issue/DEVP-624/add-persona-to-agent-settings-adk-is-blind-to-the-new-merged-role)

On the wire the merged Role field is a new `persona` component on `agentSettings`. ADK had no persona resource at all, so `pull` dropped the field builders actually author, `push` could only write the dead settings, and `diff` across environments silently ignored the primary identity field.

## Changes

- `SettingsPersona` in `resources/agent_settings.py` — plain-text `agent_settings/persona.txt`, modelled on `SettingsRules`, registered via `@register_resource("persona")`
- References validated against `{{attr:}}` and `{{vrbl:}}` — the same two the personality and role `custom` fields accepted. Attribute references travel in the content and are not tracked on the resource, since `PersonaReferences` is variables-only; so were `PersonalityReferences` and `RoleReferences`, so this is not new
- `update_persona` always carries a `variables` map (possibly empty) — the backend schema requires the key whenever `references` is sent
- `from_projection` reads `persona.content`, not the `persona` object — the projection always carries the object, so absent content just means there is nothing to write to disk. An authored empty string is a real persona and is kept
- **`SettingsPersonality` and `SettingsRole` deleted.** They still exist on the wire but nothing surfaces them, so keeping them registered only leaves two files that pull, push and validate while affecting nothing
- **Migration `removed_personality_and_role_files`** deletes `agent_settings/personality.yaml` and `role.yaml` the first time a project is loaded, logging what it removed and telling the builder to pull to get `persona.txt`. `file_structure_info` is recomputed from the loaded resources on every load, so the stale status entries drop out on their own
- Docs: the packaged `docs/agent_settings.md` + `docs/docs.md`, and the mkdocs site — the resource reference, architecture guide, working-locally tree, tooling/CLI tables and the restaurant tutorial all walked builders through `personality.yaml` and `role.yaml`

### On the compiled backfill

For a project that has never authored a persona, the content the projection returns is derived server-side from the old `personality` + `role` and stored nowhere. This is deliberately left as-is rather than detected and skipped: `push_project` only sends resources whose on-disk hash differs from the pulled snapshot, so an untouched `persona.txt` is never pushed back, and editing the file authors a real persona — which is the intent. Documented in `docs/agent_settings.md`.

## Test strategy

- [x] Manual CLI testing (`poly docs agent_settings`)

New coverage: `SettingsPersonaTests` (15 cases — raw/pretty roundtrip, attribute and variable validation, `{{fn:}}` rejection, update proto with and without references, `from_projection` for present / absent-content / absent-object / empty-string) and `TestRemovePersonalityAndRoleFiles` for the migration (no-op without `agent_settings/`, both files removed with the warning logged, other settings files untouched).

Also verified offline that a `Command(type="update_persona", ...)` builds, serialises and round-trips with `references.variables` populated.

Not yet run against a live project — `poly pull` / `poly push` on a real project is still outstanding.

## Checklist

- [x] `pytest` passes (1352 passed, 112 subtests)
- [x] No breaking changes to the `poly` CLI interface
- [x] Commit messages follow conventional commits

### DIFF
--- a/src/poly/project.py
+++ b/src/poly/project.py
@@ -40,9 +40,6 @@ from poly.resources import (
     Pronunciation,
     Resource,
     ResourceMapping,
-    SettingsPersona,
-    SettingsPersonality,
-    SettingsRole,
     SubResource,
     TestCase,
     Topic,
@@ -123,8 +120,6 @@ class AgentStudioProject:
     # So they aren't considered locally deleted when pushing/pulling
     # before they are saved.
     _not_loaded_resources: list[ResourceType] = None
-
-    push_warnings: list[str] = None
 
     @property
     def all_resources(self) -> list[Resource]:
@@ -1204,8 +1199,6 @@ class AgentStudioProject:
                 - List of commands serialized to protobuf.
         """
 
-        self.push_warnings = []
-
         if not dry_run:
             # If force, load latest version of the project
             # to compare against
@@ -1293,19 +1286,6 @@ class AgentStudioProject:
         new_resources.update(subresource_changes.new)
         updated_resources.update(subresource_changes.updated)
         deleted_resources.update(subresource_changes.deleted)
-
-        if new_state.get(SettingsPersona):
-            shadowed = [
-                resource.file_path
-                for resource_type in (SettingsPersonality, SettingsRole)
-                for resource in updated_resources.get(resource_type, {}).values()
-            ]
-            if shadowed:
-                self.push_warnings.append(
-                    f"{', '.join(sorted(shadowed))} changed, but this project uses "
-                    f"{next(iter(new_state[SettingsPersona].values())).file_path} — "
-                    "personality and role no longer affect the Role field in Agent Studio."
-                )
 
         if not (updated_resources or new_resources or deleted_resources):
             return False, "No changes detected", []
@@ -1552,10 +1532,6 @@ class AgentStudioProject:
             current_resources=self.resources,
             queue_command=lambda command: self.api_handler.queue_command(command),
         )
-
-        # There is no create_persona command; authoring a persona is an update.
-        if new_personas := new_resources.pop(SettingsPersona, None):
-            updated_resources.setdefault(SettingsPersona, {}).update(new_personas)
 
         return PushPhaseChangeSet(
             main=ResourceChangeSet(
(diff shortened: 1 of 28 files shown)


Ruari-Phipps (commented):
If personality and role are deprecated, I think we should delete them from the ADK?
(Or de-register them if we want to keep code for some reason)
That way we shouldn't have issues with people trying to push and don't need the push warnings.

Additionally, we should probably have a migration step that deletes the old files. Maybe put out a log warning that they have been migrated and need to pull to get the persona

Thirdly, can we update the main docs as well. Thanks!

src/poly/project.py
```diff
@@ -1533,6 +1553,10 @@ def _clean_resources_before_push(
             queue_command=lambda command: self.api_handler.queue_command(command),
         )
 
+        # There is no create_persona command; authoring a persona is an update.
+        if new_personas := new_resources.pop(SettingsPersona, None):
+            updated_resources.setdefault(SettingsPersona, {}).update(new_personas)
```
Ruari-Phipps: Is this needed? pre push we do a pull, that should mean it exists

src/poly/project.py
```diff
@@ -1287,6 +1294,19 @@ def push_project(
...
+                resource.file_path
+                for resource_type in (SettingsPersonality, SettingsRole)
+                for resource in updated_resources.get(resource_type, {}).values()
+            ]
+            if shadowed:
+                self.push_warnings.append(
+                    f"{', '.join(sorted(shadowed))} changed, but this project uses "
+                    f"{next(iter(new_state[SettingsPersona].values())).file_path} — "
+                    "personality and role no longer affect the Role field in Agent Studio."
+                )
```
Ruari-Phipps: I think we should just deprecate, so not even read them

src/poly/project.py
toswe: Removed.

src/poly/project.py
toswe: Removed as well.

Ruari-Phipps (approved):
Nice!

PR #304, opened 2026-09-02
feat: add AI agent skills suite

Ruari-Phipps:
## Summary

Adds a `skills/` directory with five agent skills that teach AI coding agents (Claude Code, Cursor, Codex, etc.) the `poly` CLI workflow, plus a README describing the suite. Installable via `npx skills add`.

First PR of a three-part stack: #305 adds `poly setup`, which installs these skills, and #303 makes `poly update` keep them current.

## Motivation

AI coding agents working on ADK projects currently have no contextual knowledge of the `poly` workflow — resource schemas, the no-`main` rule, pushed-state testing semantics, or conflict resolution. These skills provide that context on demand, structured after Google's `agents-cli` skills suite: a workflow entrypoint plus task-specific skills that load only when relevant.

## Changes

- `skills/poly-adk-workflow/SKILL.md` — entrypoint: install/update, auth, `poly docs` habit, resource-choice guidance, project structure, the core edit → validate → push → test → merge loop, and a routing table to the task skills
- `skills/poly-adk-testing/SKILL.md` — `poly validate`, scripted `poly chat`, `test_suite/` authoring with `api_mocks`, `poly functions execute/validate`
- `skills/poly-adk-branching/SKILL.md` — branch semantics, three-way merge model, conflict markers, non-interactive `merge --resolutions`, review gists
- `skills/poly-adk-conversations/SKILL.md` — `poly conversations`, instrumenting with `conv.log` and metrics, real-call → test-case loop
- `skills/poly-adk-rtc/SKILL.md` — RTC pull/push cycle, drift protection, live-environment safety
- `skills/README.md` — suite overview and install instructions

Resource schemas are deliberately not duplicated — skills instruct agents to run `poly docs`, which ships schemas with the installed CLI. Skill `metadata.version` is pinned to the current release (0.53.1).

## Test strategy

- [ ] Added/updated unit tests
- [x] Manual CLI testing (`poly <command>`)
- [x] Tested against a live Agent Studio project
- [x] N/A (docs, config, or trivial change)

Markdown-only change. Skills were installed locally via `npx skills add` (frontmatter validated by the tool) and exercised in Claude Code sessions against a real project. All CLI commands and flags referenced were cross-checked against `docs/` and, where relevant, the source.

## Checklist

- [x] `ruff check .` and `ruff format --check .` pass
- [x] `pytest` passes
- [x] No breaking changes to the `poly` CLI interface (or migration path documented)
[...]

PR #305, opened 2026-09-02
feat: add poly setup command and remove poly start

Ruari-Phipps:
## Summary

Adds `poly setup` — a single onboarding command covering authentication, shell completion, AI agent skill installation, and project setup — and removes `poly start`, which it supersedes. Second PR of the stack: stacked on #304 (the skills it installs), and the base for #303, which makes `poly update` keep them current.

## Motivation

Onboarding currently spans several commands users must discover one by one (`poly login`, `poly completion`, `poly project create`), and the new agent skills had no installation path. `poly setup` runs all of it in one command, skipping any step that is already done, so it doubles as a repair command. `poly start` covered a subset of this (auth + project, hardcoded to the `studio` region) and is not referenced in published material, so it is removed rather than left as a second entry point.

## Changes

- New `poly setup` command with `--region`, `--base-path`, `--skip-auth`, `--skip-skills`, `--agent` (repeatable), `--dev`, and `--global/-g` flags
- New `cli_commands/skills.py`: wrapper around the pinned `npx skills` package with a Node.js 18+ gate; skill installation is non-fatal — a missing Node warns and setup continues
- `poly login` (and setup) now wait up to 20s for a newly created API key to become active before returning
- Region selection extracted to a shared picker; `poly start`'s four hardcoded `studio` call sites removed along the way
- `poly start` removed from the CLI, docs, and nav
- Docs: new `setup` reference page; getting-started restructured to "install, then `poly setup`"; tooling page now leads with skill installation via setup
- `poly-adk-workflow` skill updated to mention `poly setup` for fresh machines
- Root README quickstart updated from `poly start` to `poly setup`

## Test strategy

60 new tests (node gate, npx argument construction and non-raising failures, per-step skip logic, real rc-file completion installs in throwaway home directories — set via both HOME and USERPROFILE so they are hermetic on Windows — region threading, and activation-poll behavior). Manually verified end-to-end with an isolated `$HOME`: completion install + idempotent re-run, `--dev -g --agent claude-code` installing all five skills via npx, graceful skip without a TTY, and `poly start` now reporting an invalid choice.

## Checklist

- [x] `pytest` passes
- [x] No breaking changes to the `poly` CLI interface (or migration path documented) — `poly start` is intentionally removed; `poly setup` / `poly login` are the migration path

## Screenshots / Logs

N/A


jamesosullivan-polyai:
## Review: blocking issue in project detection

`SetupCommand._setup_project` detects an existing project with:

```python
if os.path.exists(os.path.join(base_path, "project.yaml")):
    success("Project: already inside an ADK project — skipping.")
    return
```

This diverges from the project's own source of truth for "is this a project", `read_project_config` in [`shared.py`](../blob/main/src/poly/cli_commands/shared.py#L19-L47), which every other command (`status`, `push`, `diff`, ...) uses. That function checks `project.yaml` **and** falls back to the `_gen/.agent_studio_config` status file — `AgentStudioProject.from_dict()` can fully reconstruct a project from the status file alone, and this dual-file design dates back to the initial commit, so it's a deliberately supported state, not legacy cruft.

If a directory ever ends up with the status file but no `project.yaml` (e.g. a `.gitignore` that doesn't track `project.yaml`, or a partial restore), `poly setup` will report "no project here" and offer to create/connect a new one, even though every other command in the tool treats the directory as a fully working project. That undercuts the "safe to re-run" idempotency this command is designed around.

Suggest reusing `read_project_config(base_path)` (or at least checking both `PROJECT_CONFIG_FILE` and `STATUS_FILE` via the shared constants) instead of the hardcoded, single-file check.

Minor, non-blocking while I'm here:
- Same function: prefer `poly.project.PROJECT_CONFIG_FILE` over the literal `"project.yaml"` to avoid drift if the filename ever changes.
- `poly start` is removed with no alias/shim — reasonable per the PR description, but worth a release-notes callout for anyone with it in scripts.
- `setup_test.py::SigninDeviceFlowTest` docstring still references the now-removed `start` command.

— Claude

jamesosullivan-polyai (approved):
Approving per author's request. Flagging again for visibility: _setup_project's project-detection check (see comment above) diverges from read_project_config's project.yaml-or-status-file fallback used by every other command, which could make `poly setup` misreport an already-working project as "no project here" in edge cases. Recommend addressing before or shortly after merge, along with the stale docstring in setup_test.py::SigninDeviceFlowTest referencing the removed `start` command.

PR #307, opened 2026-09-03
feat: support typed variant attributes

andreitelerez-poly:
## Summary

Variant attributes can now declare a type. ADK reads an attribute's `kind` (and an
enum's `config`) from the platform, validates every value against it locally, and
dual-writes native typed values on push — matching the `AttributeType` support the
platform added in the Variant Attribute Types project.

## Motivation

The platform now stores variant attribute values twice: the legacy stringified
`values` map and a native `typedValues` map that wins on read. ADK knew about
neither, so it flattened every typed attribute to a string on pull, and its pushes
were rejected outright by the backend's type validation once a project had a typed
attribute.

Linear project: https://linear.app/poly-ai/project/variant-attribute-types-a7bc31d07a9c/overview

## Changes

- Variant attributes gain an optional `kind` (`string`, `number`, `boolean`, `enum`,
  `object`) and, for enums, a `config` listing the allowed values.
- Push dual-writes both maps and sends the declared type on create and update.
- Pull prefers `typedValues`, falls back to the whole `values` map when it is
  malformed, and parses a legacy string against the declared kind — so an attribute
  typed in Agent Studio but never re-saved still reads natively.
- Values are validated against their kind before push, naming the variant a
  mismatch came from instead of surfacing it on a live call.
- `values` stays a map of strings on disk, with `kind` saying how to read them. An
  ADK released before typed attributes calls `.strip()` on every value and would
  fail on a native int, bool or nested map in a file a newer ADK had written. Native
  YAML is still accepted on input, and both spellings hash identically so neither
  shows as a diff against the other.
- New variants seed typed attributes with a native null rather than `""`, which the
  platform rejects for any non-string kind.
- Docs updated, including two behaviours that aren't discoverable from the file: a
  blank typed attribute is omitted from the deployed agent (a blank string
  substitutes empty), and Agent Studio's Date & time / Opening hours / Voice types
  are stored as `string` or `enum` here.

A project that never adopts a type sees no change to its file at all.

## Test strategy

1585 tests pass. New coverage pins the wire round trip per kind, the projection read
including the malformed-`typedValues` fallback, YAML round trips for every kind, and
the backward-compatible encoding — including a test that runs the pre-types reader's
exact comprehension over what this writes, plus a guard asserting that same reader
really does fail on a native value.

Not yet exercised against a live project. Worth a real `poly push` of a typed
attribute before merging, since this changes the push payload.

## Checklist

- [x] `pytest` passes

### DIFF
--- a/docs/docs/reference/resources/variants.md
+++ b/docs/docs/reference/resources/variants.md
@@ -46,9 +46,56 @@ Each attribute includes:
 | Field | Description |
 |---|---|
 | `name` | Attribute identifier, ideally in snake_case |
-| `values` | Map from variant name to string value |
+| `kind` | Optional. The attribute's type: `string` (default), `number`, `boolean`, `enum`, or `object` |
+| `config` | Optional. Type-specific configuration. Only `enum` takes one: `config.values` lists the allowed values |
+| `values` | Map from variant name to value |
 
-Every attribute must provide a value for every defined variant, even if that value is an empty string.
+Every attribute must provide a value for every defined variant, even if that value is blank.
+
+## Attribute types
+
+An attribute with no `kind` is a string attribute, which is how every attribute behaved before types existed — existing files need no change.
+
+Declaring a `kind` does two things: values are checked against it when you push, so a typo is caught at authoring time rather than on a live call, and the agent receives the value in its real type. A `number` attribute arrives as `3`, not `"3"`, and a `boolean` as `true`, not `"True"`.
+
+| `kind` | Written in `values` as | Example |
+|---|---|---|
+| `string` | Text | `London Office` |
+| `number` | A number | `"3"`, `"2.5"` |
+| `boolean` | `"true"` or `"false"` | `"true"` |
+| `enum` | One of `config.values` | `premium` |
+| `object` | A JSON object or array | `'{"currency":"USD"}'` |
+
+A blank value is allowed for every kind and means the attribute is not set for that variant yet.
+
+!!! note "Values are stored as text"
+
+    `values` always holds strings on disk, and `kind` says how to read them. That keeps the file readable by an ADK released before typed attributes, which would otherwise fail on a bare `3` or `true`.
+
+    You can still write values naturally by hand — `max_retries: 3` and `max_retries: "3"` are the same attribute, and neither shows as a change against the other. `poly pull` rewrites them in the quoted form.
+
+Changing an attribute's `kind` does not convert its existing values. Update the values in the same change, or the push is rejected with the variants whose values no longer fit.
+
+### Blank values at runtime
+
+A blank value behaves differently depending on the kind, and the difference only shows up on a live call:
+
+- A blank **`string`** attribute substitutes an empty string, so `{{attr:name}}` resolves to nothing.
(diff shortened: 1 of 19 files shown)


src/poly/resources/variant_attributes.py
```diff
@@ -194,27 +383,57 @@ def discover_resources(base_path: str) -> list[str]:
...
+        # rewrite the file for every project that has never used a type.
+        if self.kind != AttributeKind.STRING:
+            yaml_dict["kind"] = self.kind.value
+        if self.config:
+            yaml_dict["config"] = dict(self.config)
+        # Values are written as strings — the same strings the platform keeps in its
+        # legacy `values` map, with `kind` saying how to read them. Writing them
+        # natively would be prettier, but an ADK released before typed attributes
+        # calls .strip() on every value and would die on an int, bool or nested map
+        # the moment it read a file a newer ADK had written.
```
Ruari-Phipps: We don't have to maintain compatibility like this.

src/poly/tests/project_test.py
Ruari-Phipps, line 1: Could we also update the test project in in tests please

Ruari-Phipps (commented):
Overall looks good!

There's quite a few verbose claude comments that could be cleaned up and would be good to add this in the test project included.

Would also be good to do manual QA of this, if you haven't already

docs/docs/reference/resources/variants.md
```diff
@@ -86,6 +133,29 @@ attributes:
...
+    kind: boolean
+    values:
+      new_york: "true"
+      london: "true"
+      tokyo: "false"
+
+  - name: tier
+    kind: enum
+    config:
+      values: [basic, premium]
```
Ruari-Phipps, line 154: Would there be anything else in this field potentially in the future? Just wondering if values should be top level?

src/poly/resources/variant_attributes.py
```diff
@@ -385,6 +656,66 @@ def validate(self, resource_mappings: list[ResourceMapping], **kwargs):
                 f"Missing variants for variant attribute: {[known_variant_id_to_name[variant_id] for variant_id in missing_variants]}"
             )
```
Ruari-Phipps, line 635: When testing I got this error:
```
ERROR:poly.handlers.sync_client:Failed to send commands: Error: {'name': 'ZodValidationError', 'message': '[command 1: variantCreateAttribute "New enum" (VARIANT_ATTRIBUTES-ba66e2d9)] Validation error: Invalid value, name should be Pythonic at "name"', 'details': [{'validation': 'regex', 'code': 'invalid_string', 'message': 'Invalid value, name should be Pythonic', 'path': ['name']}]}
```
This is outside the scope of this PR, but would be good to include it in the validation while we're here

Ruari-Phipps (commented):
Bug I also found.
1. Push enum variant attribute with a config
2. Change attribute type and push
3. Pull -> config returns and it fails validation

Either:
- Don't write config if not enum type
- Or we need to correctly clear the config when pushed empty. This might require a backend change

docs/docs/reference/resources/variants.md
andreitelerez-poly: No plans to extend it for this feature. `enum` is the only kind with metadata today. It's nested so a future kind can carry its own settings without every other kind growing a field it ignores.

Ruari-Phipps (approved):

PR #314, opened 2026-09-09
feat: proactively match parent branch ids on push

Ruari-Phipps:
## Summary

`poly push` now proactively adopts the parent branch's resource ids for new local resources that match by file path, and mints subresource ids (conditions, function parameters, delay responses) deterministically from their scoped names. This keeps ids aligned across branches so merges reconcile resources instead of duplicating them.

## Motivation

When a resource is added on the parent branch and its files reach a child branch (e.g. via a git-side merge of the project repo), pushing previously minted fresh random ids for them. The same logical resource then had a different id on each branch, and merges treated the two copies as unrelated — producing duplicates or name collisions. Adopting existing ids at mint time and deriving subresource ids from stable names removes the main sources of this divergence.

## Changes

- `poly push` fetches the parent branch's resources by file path and passes them into `find_new_kept_deleted`; a new resource whose file path-matches a parent resource adopts the parent's id at mint time. Flows resolve first so composite step ids stay consistent, and steps adopt the parent's bare step id under the local flow prefix.
- Parent-id adoption always runs on real pushes; dry runs skip the parent fetch unless `POLY_ADK_SYNC_PARENT_IDS_TEST=1` forces it for testing.
- Kept resources inherit parent-only subresources by name, so a condition or parameter added on both the parent and the branch ends up with one shared id.
- Condition, function parameter, and delay-response ids are derived from their scoped names (`generate_subresource_id`) instead of minted randomly — repeated reads are idempotent and independently created same-named subresources agree across branches.
- `sync_ids_with_sandbox` is generalized to `sync_ids_with_parent`, defaulting to the branch's actual parent and falling back to main.
- The parent branch's projection can be supplied directly — `push_project(parent_projection_json=...)` for library consumers (e.g. a service embedding the ADK), mirrored by a hidden `--parent-projection JSON|-` CLI flag. A supplied projection builds the parent-id lookup fully offline (no branch or projection fetch) and enables adoption on dry runs; an empty dict means "no parent"; omitting it keeps the fetch behavior.
- Parent fetching and path-keying are split (`_fetch_parent_resources` returning the raw resources + a shared `_resources_by_absolute_path`), and `sync_ids_with_parent` reuses the same fetch for its default-parent resolution.
- Retargeted the id-sync tests at the mint site and added coverage for deterministic ids, subresource inheritance, the offline parent-projection path (enforced by a test API handler that fails on any network call), and the new CLI parsing.

## Test strategy

Manual test pass:

- [x] 1. New resource from the parent adopts the parent's id on push
- [x] 2. New flow with steps — flow id adopted, step ids bare in commands, conditions carry parent ids, references resolve
- [x] 3. New step in a kept flow adopts the parent's bare step id under the branch flow prefix
- [x] 4. Condition added on both sides keeps the parent's condition id (kept step)
- [x] 5. Parameter added on both sides keeps the parent's parameter id
- [x] 6. Repeated dry runs mint identical condition ids (read idempotence)
- [x] 7. Sibling branches mint identical ids for the same-named condition
- [x] 8. Delay-response ids stable across pushes
- [x] 9. Push from main unchanged
- [x] 10. Dry run without the test env var does not fetch the parent
- [x] 11. Local-only resource still mints a random id
- [x] 12. Parent fetch failure mode on a real push
- [x] 13. `sync_ids_with_parent` against a non-main parent
- [x] 14. Pull + status clean after a push with adopted ids
- [x] 15. Merge with an untouched adopted-id resource auto-resolves (audit-only diff)
- [x] 16. Merge with a branch-edited adopted-id resource surfaces a real conflict
- [x] 17. Merge without the audit-field auto-resolve deployed (control)
- [x] 18. `poly push --dry-run --parent-projection -` (projection piped in) shows adopted ids fully offline

## Checklist

- [x] `pytest` passes
- [x] No breaking changes to the `poly` CLI interface (or migration path documented) — note: real pushes now always attempt parent-id adoption for new resources; dry-run output is unchanged unless the test env var is set

### DIFF
--- a/src/poly/cli.py
+++ b/src/poly/cli.py
@@ -190,7 +190,7 @@ class AgentStudioCLI:
                     command.run(args)
                     return
         except Exception as e:
-            if hasattr(args, "json") and args.json:
+            if getattr(args, "json", False) or getattr(args, "output_json_commands", False):
                 json_print({"success": False, "error": str(e), "traceback": traceback.format_exc()})
                 sys.exit(1)
             else:
--- a/src/poly/cli_commands/shared.py
+++ b/src/poly/cli_commands/shared.py
@@ -173,10 +173,16 @@ def parse_from_projection_json(
     from_projection: Optional[str],
     *,
     json_errors: bool,
+    flag_name: str = "--from-projection",
 ) -> Optional[dict[str, Any]]:
-    """Parse ``--from-projection`` CLI value into a projection dict, or exit on failure.
+    """Parse a projection-carrying CLI value into a projection dict, or exit on failure.
 
     If the value is ``-`` (after stripping), JSON is read from stdin until EOF.
+
+    Args:
+        from_projection (Optional[str]): The raw CLI value; None/empty returns None.
+        json_errors (bool): Emit failures as JSON instead of console errors.
+        flag_name (str): The flag the value came from, used in error messages.
     """
     from poly.output.console import error
 
@@ -190,14 +196,14 @@ def parse_from_projection_json(
         if isinstance(parsed, dict) and "projection" in parsed:
             parsed = parsed["projection"]
     except json.JSONDecodeError as e:
-        msg = f"Invalid JSON in --from-projection: {e}"
+        msg = f"Invalid JSON in {flag_name}: {e}"
         if json_errors:
             json_print({"success": False, "error": msg})
         else:
             error(msg)
         sys.exit(1)
     if not isinstance(parsed, dict):
-        msg = "--from-projection must be a JSON object (dictionary)."
+        msg = f"{flag_name} must be a JSON object (dictionary)."
         if json_errors:
             json_print({"success": False, "error": msg})
         else:
--- a/src/poly/cli_commands/sync.py
+++ b/src/poly/cli_commands/sync.py
@@ -327,6 +327,13 @@ class PushCommand(BaseCommand):
             default=None,
         )
         push_parser.add_argument(
+            "--parent-projection",
+            type=str,
+            metavar="JSON|-",
+            help=SUPPRESS,
+            default=None,
+        )
+        push_parser.add_argument(
             "--output-json-commands",
             action="store_true",
             help=SUPPRESS,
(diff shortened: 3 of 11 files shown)


jamesosullivan-polyai:
Two things from a read-through — neither blocking.

**1. Delay-response ids can collide within a single function**

`generate_subresource_id("DELAY", msg, str(dur))` derives the id from message + duration only, so two entries with the same message and duration in one `@func_latency_control` now get the identical id:

```python
@func_latency_control(delay_responses=[("One moment.", 2), ("One moment.", 2)])
```

Both resolve to the same `DELAY-xxxxxxxx`. The `used_delay_response_ids` set just above (`function.py` L944-955) exists specifically to stop duplicate-message entries from reusing a single known id, so the code already treats same-message duplicates as a real scenario — the new fallback reintroduces the collision that guard was written for. Worth noting that adding the function name to the scope wouldn't fix this one; it needs something occurrence-specific (e.g. the index within the list) in the scope.

Separately, for consistency: `CONDITION` and `PARAMETER` are scoped by their full hierarchy (flow / step / name, function / name) but `DELAY` isn't scoped by its function at all. Harmless today since delay ids only ever travel nested under a function-scoped payload, but it's an odd asymmetry to leave in.

**2. Parent fetch failure on a real push**

`_fetch_parent_resources()` is unconditional on real pushes and isn't guarded. The no-parent cases are all handled and covered by tests (main, no recorded parent, branch gone from the remote), and a plain network outage would already have failed at the `pull_project()` a few lines above — so this is narrow rather than a general reliability concern.

The case I'm less sure about is the parent branch specifically being unreadable while the current branch is fine: parent deleted but `parentBranchId` still recorded on the child, or per-branch permissions differing. If `pull_resources()` raises there, the push aborts *after* the pull has already written to the working tree. Did item 12 in your manual test pass cover the deleted-parent case, or just an unreachable fetch?

Rest looks good — test coverage on the id-adoption paths is thorough, particularly the cross-flow/step name-collision cases and the network-refusing fake handler pinning the offline projection path.

jamesosullivan-polyai (approved):

PR #315, opened 2026-09-10
feat: add `poly apikey` for one-shot GitHub sign-in and API key setup

andymohajeri:
## Summary

Adds `poly apikey`, a non-interactive command whose job is to get an API key for the PolyAI
APIs and Dialog RSN into a user's environment - not to set up the ADK itself, which stays with
`poly setup`/`poly login`. It signs a user in through the device flow, creates their Agent
Studio account if needed, mints (or reuses) an account-scoped API key, saves it to the
credentials file, and exports `POLY_API_KEY` into the user's shell profile (or user environment
on Windows). `--region studio` targets the PLG cluster via a GitHub-only sign-in page, for
individual developers; `--region <region>` targets any other cluster (us-1/uk-1/euw-1) via the
same email/SSO page `poly login` uses, for enterprise users who need a key for a region other
than where their identity lives. It's runnable by an AI coding assistant on the user's behalf,
since it never prompts and handles no secrets outside the process.

## Motivation

The studio sign-in page is getting a "Set up with your AI coding assistant" panel: a prompt the
user copies into Claude Code, Cursor or Codex. The first version of that prompt had the agent
drive the Auth0 device flow and edit dotfiles directly. Agents frequently refused it as a
prompt-injection pattern, and when they did run it the token and key passed through the agent's
transcript. Moving the flow into the CLI makes the pasted prompt "install the package and run one
command", which agents accept, and keeps every secret inside the process. `poly apikey` is
scoped narrowly to the API-key/Dialog RSN use case; ADK onboarding for an assistant still goes
through `poly setup`/`poly login`, unchanged by this PR.

Sign-up tracking is done on the Auth0 side (post-login Action -> PostHog), so the ADK sends no
telemetry.

Why this is a separate command rather than a flag on `poly login`, or part of `poly setup` (#305):

- It authenticates against a different Auth0 application, whose login page shows only
  "Continue with GitHub". `login` and `setup` keep the email/Google/GitHub page.
- It creates an **account-scoped** API key on the accounts API, not a user PAT.
- It exports `POLY_API_KEY` for use by the PolyAI SDKs and other tools, which `login`
  deliberately does not. The ADK itself keeps reading `~/.poly/credentials.json`, so the variable
  is for everything downstream of onboarding rather than for `poly` commands.
- It never prompts, and its `--json` shape is a stable contract for the agent driving it.
  `setup` is the interactive, human-facing path and stays that way.

## Changes

- Add `poly apikey` (Getting started group) with `--region`, `--key-name`, `--account-id`,
  `-f/--force`, and the shared `--json` / `--verbose` / `--debug` parents. Exit `0` on success,
  `1` on sign-in, account or API failure, `2` when `POLY_API_KEY` is already set to a different
  value and `--force` was not given. After saving credentials it polls (up to 20×, 1 s apart,
  mirroring `poly login`/`poly setup`) for the key to become active before writing
  `POLY_API_KEY`, so a command run immediately afterwards doesn't fail against a key the
  platform hasn't finished activating; if it never activates, it warns and continues rather
  than failing.
- `--region` (required, no default) picks which cluster's Auth0 app to sign in against and which
  cluster's account/key APIs to call - reusing `REGION_TO_AUTH_DETAILS` and the region-keyed
  interface methods `poly login` already uses, so no new Auth0 infrastructure was needed. The
  default key name becomes `cli-generated-key-<region>` for any non-`studio` region, so a key
  created for one region never collides with or overwrites a key from another - each region's
  credential also gets its own independent entry in `~/.poly/credentials.json`.
- Extract the RFC 8628 device-flow loop from `auth.py` into `poly.auth.device_flow`, taking an
  `AuthDetails` rather than a region so a caller can target a specific Auth0 application. `login`
  and `setup` keep their exact output; they now call the shared helper. The poll interval is
  capped at 30 s under repeated `slow_down`.
- Add JWT-authenticated platform calls for listing accounts and listing/creating account API
  keys, and factor the repeated Bearer header block into `_jwt_headers(jwt, source)`. The
  `X-Poly-Source` header is parameterised so apikey traffic is distinguishable from the rest
  of the ADK; existing callers still send `adk`.
- Add `poly.utils.api_keys.select_reusable_api_key`: newest active, unexpired key whose
  `policies[0].name` matches (the creation name lands there, not on the key's own `name`).
- Add `poly.utils.env_profile`: detects the profile file from `$SHELL` (`~/.zshrc`,
  `~/.bash_profile` on macOS / `~/.bashrc` on Linux, fish config, `~/.profile` fallback) and
  writes the export line with a marker comment, replacing an existing line in place only with
  `--force`. On Windows it writes the user-scope registry environment and broadcasts
  `WM_SETTINGCHANGE`, via a small backend seam so the logic is testable on Linux CI.
- Never overwrite an existing `studio` entry in `~/.poly/credentials.json`; report that it was
  kept.
- On enterprise regions (`us-1`/`uk-1`/`euw-1`), `/jupiter/v1/authorise` only auto-creates a
  workspace for `@poly-ai.com` users (verified in `poly_core`,
  `src/agent_v3/jupiter_api/interactors/auth_interactor.py`) - anyone else who isn't already
  provisioned in Aegis gets a raw `403`. `apikey` now catches that and raises a friendly
  message pointing at `--account-id`/PolyAI provisioning instead of surfacing the HTTP error;
  studio's 403 behaviour (a genuine auth error) is untouched. Also: account resolution now
  refuses to silently pick among multiple accounts on any region - including studio - erroring
  and listing them unless `--account-id` is given; the sign-in message no longer reads "with
  your browser ... in your browser" for non-studio regions, and `--json` output includes
  `key_active` so an agent can tell whether the activation wait actually succeeded.
- Docs: reference page for `apikey`, entries in the CLI overview and nav, a short subsection in
  Getting started, and a README line.

## Test strategy

Unit tests (run on both the Ubuntu and Windows CI runners) cover the device flow (first-poll success, pending, slow-down cap, expiry, unknown
error, no-browser, custom callback), key selection (policy-name match, inactive, expired, missing
expiry, newest wins), profile detection and writing for zsh/bash/fish/fallback and the Windows
registry path (fake backend), and the command itself (create and reuse paths, `--account-id`,
empty-account timeout, conflict exit 2, `--force` forwarding, `--json` shape, key never printed,
region selection - `--region studio` vs. `--region us-1` sign-in target, per-region default key
name, `--key-name` overriding it, region threaded into account/credentials calls, `--region`
required). Tests make no network calls and do not touch the real `~/.poly`.

Manual runs against studio with a fresh GitHub account, all passing end to end:

| Platform | Shell / store | Paths exercised |
|---|---|---|
| macOS 15 | zsh, `~/.zshrc` | new account, key created, conflict exit 2, `--force` replace in place |
| Debian 12 (Docker) | bash, `~/.bashrc` | key reused, append with marker, fresh login shell sees the variable |
| Windows Server 2022 (GCP) | user registry | key reused, `HKCU\Environment` written, value visible from a new process |

In each case `poly project list --region studio` authenticated with the saved credential
afterwards.

`--region` for a non-`studio` cluster has not been exercised against a live enterprise tenant in
this environment (no credentials for one here) - it reuses the exact `AuthDetails`/device-flow/
interface calls `poly login --region <region>` already uses in production, so risk is limited to
this file's own plumbing, covered by the unit tests above.

## Checklist

- [x] `pytest` passes

`--region` went from optional (defaulting to `studio`) to required in a later push of this same
PR. `poly apikey` has never shipped in a release, so this is not a break in released behaviour -
just tightening the interface before it goes out.

Overlaps with #305 (`poly setup`) in `auth.py` imports, `cli.py`'s command list and the docs
index pages; whichever lands second rebases. This branch no longer references `poly start`, so
merge order does not change its docs.

## Screenshots / Logs

```
$ poly apikey --region studio
Setting up your PolyAI account and API key...
To sign in with GitHub, open the following link in your browser
and enter the code when prompted.

  URL:  https://login.studio.poly.ai/activate?user_code=XXXX-XXXX
  Code: XXXX-XXXX
Authenticated successfully!
Setting up your account...
Created API key 'cli-generated-key': abcd****wxyz
Saved to ~/.poly/credentials.json (studio).
Wrote to ~/.zshrc:
  export POLY_API_KEY="abcd****wxyz"

Done.
  Account:   ws-xxxxxxxx
  API key:   abcd****wxyz   (cli-generated-key, created)
  Saved to:  ~/.poly/credentials.json (studio)
  Profile:   ~/.zshrc  (export POLY_API_KEY="abcd****wxyz")

Open a new terminal, or run `source ~/.zshrc`, to pick up POLY_API_KEY.
```

### DIFF
--- a/docs/docs/get-started/get-started.md
+++ b/docs/docs/get-started/get-started.md
@@ -79,12 +79,12 @@ poly login --region us-1
 
 To sign in to more than one region from the same machine, re-run `poly login` for each — the credential file stores them side by side.
 
-### Setup from an AI coding assistant — `poly onboard`
+### Setup from an AI coding assistant — `poly apikey`
 
-If an AI coding agent is setting this up on your behalf, it can run [`poly onboard`](../reference/cli/onboard.md) — a non-interactive, one-shot command that signs in via GitHub, provisions an account-scoped API key, and exports `POLY_API_KEY` into your shell profile without any prompts.
+If an AI coding agent is setting this up on your behalf, it can run [`poly apikey`](../reference/cli/apikey.md) — a non-interactive, one-shot command that signs in via GitHub, provisions an account-scoped API key, and exports `POLY_API_KEY` into your shell profile without any prompts.
 
 ```bash
-poly onboard
+poly apikey
 ```
 
 ### Manual API key export { #manual-api-key-export }
--- a/docs/docs/get-started/get-started.md
+++ b/docs/docs/get-started/get-started.md
@@ -79,6 +79,14 @@ poly login --region us-1
 
 To sign in to more than one region from the same machine, re-run `poly login` for each — the credential file stores them side by side.
 
+### Setup from an AI coding assistant — `poly onboard`
+
+If an AI coding agent is setting this up on your behalf, it can run [`poly onboard`](../reference/cli/onboard.md) — a non-interactive, one-shot command that signs in via GitHub, provisions an account-scoped API key, and exports `POLY_API_KEY` into your shell profile without any prompts.
+
+```bash
+poly onboard
+```
+
 ### Manual API key export { #manual-api-key-export }
 
 If you would rather store your credentials in an environment variable - in a CI for example - create the key yourself in the Agent Studio UI:
--- a/docs/docs/get-started/get-started.md
+++ b/docs/docs/get-started/get-started.md
@@ -79,12 +79,12 @@ poly login --region us-1
 
 To sign in to more than one region from the same machine, re-run `poly login` for each — the credential file stores them side by side.
 
-### Setup from an AI coding assistant — `poly apikey`
+### Need an API key for the PolyAI APIs or Dialog RSN? — `poly apikey`
 
-If an AI coding agent is setting this up on your behalf, it can run [`poly apikey`](../reference/cli/apikey.md) — a non-interactive, one-shot command that signs in via GitHub, provisions an account-scoped API key, and exports `POLY_API_KEY` into your shell profile without any prompts.
+`poly apikey --region studio` signs you in, creates an account-scoped API key, and exports `POLY_API_KEY` into your shell profile for use in your own code. It's non-interactive, so an AI coding assistant can run it for you. For setting up the ADK itself, `poly setup`/`poly login` above is the path — the ADK doesn't need `POLY_API_KEY`.
 
 ```bash
-poly apikey
+poly apikey --region studio
 ```
 
 ### Manual API key export { #manual-api-key-export }
--- a/docs/docs/reference/cli/apikey.md
+++ b/docs/docs/reference/cli/apikey.md
@@ -60,9 +60,5 @@ Install [uv](https://docs.astral.sh/uv/) with `winget install astral-sh.uv`. If 
 
 With `--json`, the sign-in URL and code are printed to stderr instead of being suppressed, since the JSON contract only constrains stdout and an agent that can't open a browser still needs them to complete sign-in. `key_active` is `false` if the key never finished activating within the poll window — the human-readable warning that would normally say so is suppressed in `--json` mode, so this is how an agent detects it.
 
-## Telemetry
-
-`poly apikey` sends anonymous usage events (`apikey_started`, `apikey_authenticated`, `apikey_account_resolved`, `apikey_key_reused`/`apikey_key_created`, `apikey_env_written`, `apikey_completed`, `apikey_failed`) to help improve the flow; the API key and sign-in token are never included. Set `DO_NOT_TRACK=1` or `POLY_NO_TELEMETRY=1` to opt out.
-
 !!! info "Designed for AI coding assistants"
     `poly apikey` never prompts, so it can be run unattended by an agent. It is not a replacement for [`poly login`](./login.md) — that remains the interactive path for a human at the keyboard.
--- a/docs/docs/reference/cli/apikey.md
+++ b/docs/docs/reference/cli/apikey.md
@@ -19,8 +19,8 @@ poly apikey --region us-1
 `poly apikey`:
 
 1. Signs in via the Auth0 device authorization flow. For the default `studio` region, this is a dedicated client whose login page shows only "Continue with GitHub". For any other `--region`, it's the same standard sign-in page (email/SSO) that [`poly login`](./login.md) uses.
-2. Calls the authorise endpoint, which creates your account in that region if it doesn't already exist.
-3. Resolves your account id — by default, polls for up to 20 seconds for the newly created account to appear; pass `--account-id` to skip this.
+2. Calls the authorise endpoint. On `studio`, this creates your workspace if you don't already have one. On an enterprise region (`us-1`, `uk-1`, `euw-1`), your account must already be provisioned by PolyAI — the command stops with a clear message if it isn't.
+3. Resolves your account id — by default, polls for up to 20 seconds for the newly created account to appear; pass `--account-id` to skip this. On an enterprise region, more than one account without `--account-id` is an error rather than a guess: the command lists the accounts found and asks you to pick one.
 4. Reuses an active, unexpired API key with the default name if one already exists for the account, otherwise creates one. Unlike [`poly login`](./login.md)'s Personal Access Token, this key is **account-scoped**.
 5. Saves the key to `~/.poly/credentials.json` under `--region` — but only if no entry for that region already exists there. An existing ADK credential is never overwritten, and credentials for different regions coexist independently.
 6. Writes `export POLY_API_KEY="..."` (or, on Windows, sets the variable in your user environment) into your detected shell profile, so new shells and processes pick it up automatically. The ADK itself reads `~/.poly/credentials.json` and doesn't need this variable; it's exported for the PolyAI SDKs and other tools that read `POLY_API_KEY` directly.
@@ -53,11 +53,12 @@ Install [uv](https://docs.astral.sh/uv/) with `winget install astral-sh.uv`. If 
   "api_key_masked": "sk-n****7890",
   "credentials_file": "/home/user/.poly/credentials.json (studio)",
   "profile_path": "/home/user/.zshrc",
-  "profile_shell": "zsh"
+  "profile_shell": "zsh",
+  "key_active": true
 }
 ~~~
 
-With `--json`, the sign-in URL and code are printed to stderr instead of being suppressed, since the JSON contract only constrains stdout and an agent that can't open a browser still needs them to complete sign-in.
+With `--json`, the sign-in URL and code are printed to stderr instead of being suppressed, since the JSON contract only constrains stdout and an agent that can't open a browser still needs them to complete sign-in. `key_active` is `false` if the key never finished activating within the poll window — the human-readable warning that would normally say so is suppressed in `--json` mode, so this is how an agent detects it.
 
 ## Telemetry
 
--- a/docs/docs/reference/cli/apikey.md
+++ b/docs/docs/reference/cli/apikey.md
@@ -5,22 +5,22 @@ description: Reference for the `poly apikey` command.
 
 # `poly apikey`
 
-One-shot setup for an AI coding assistant: sign-in, an account-scoped API key, and `POLY_API_KEY` exported into your shell profile. Never prompts — use [`poly login`](./login.md) for the interactive, human-driven flow.
+`poly apikey` is the narrower cousin of [`poly login`](./login.md): it gets you an account-scoped API key for the PolyAI APIs and Dialog RSN, not an ADK setup. It signs you in, creates (or reuses) that key, and exports it as `POLY_API_KEY` into your shell profile. Never prompts, so an AI coding assistant can run it for you. To set up the ADK itself, use [`poly setup`](./setup.md) or [`poly login`](./login.md) instead — the ADK reads its credentials from `~/.poly/credentials.json` and does not need `POLY_API_KEY`.
 
 Examples:
 
 ~~~bash
-poly apikey
-poly apikey --account-id acc-123
-poly apikey --key-name my-key --force
+poly apikey --region studio
+poly apikey --region studio --account-id acc-123
+poly apikey --region studio --key-name my-key --force
 poly apikey --region us-1
 ~~~
 
 `poly apikey`:
 
-1. Signs in via the Auth0 device authorization flow. For the default `studio` region, this is a dedicated client whose login page shows only "Continue with GitHub". For any other `--region`, it's the same standard sign-in page (email/SSO) that [`poly login`](./login.md) uses.
+1. Signs in via the Auth0 device authorization flow. For `--region studio`, this is a dedicated client whose login page shows only "Continue with GitHub". For any other `--region`, it's the same standard sign-in page (email/SSO) that [`poly login`](./login.md) uses.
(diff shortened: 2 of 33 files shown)


docs/docs/get-started/get-started.md
```diff
@@ -79,6 +79,14 @@ poly login --region us-1
...
 To sign in to more than one region from the same machine, re-run `poly login` for each — the credential file stores them side by side.
 
+### Setup from an AI coding assistant — `poly apikey`
+
+If an AI coding agent is setting this up on your behalf, it can run [`poly apikey`](../reference/cli/apikey.md) — a non-interactive, one-shot command that signs in via GitHub, provisions an account-scoped API key, and exports `POLY_API_KEY` into your shell profile without any prompts.
+
+```bash
+poly apikey
+```
+
```
Ruari-Phipps: We should talk about how this for use for using the APIs or Dialog RSN. AI coding assistants wanting to use the ADK should just go through `poly login` or `poly setup`

docs/docs/reference/cli/apikey.md
```diff
@@ -0,0 +1,68 @@
+---
+title: poly apikey
+description: Reference for the `poly apikey` command.
+---
+
+# `poly apikey`
+
+One-shot setup for an AI coding assistant: sign-in, an account-scoped API key, and `POLY_API_KEY` exported into your shell profile. Never prompts — use [`poly login`](./login.md) for the interactive, human-driven flow.
```
Ruari-Phipps: Same as above

docs/docs/tooling/tooling.md
```diff
@@ -72,6 +72,8 @@ Reference `rules.md` in your session prompt. This gives the coding tool accurate
 
     The IDE extension and Claude Code cover different modes of work. You can edit in VS Code or Cursor day-to-day and still reach for Claude Code when you want an agent to generate or refactor a large slice of the project on your behalf.
 
+If you are setting up the ADK from Claude Code or another coding assistant, [`poly apikey`](../reference/cli/apikey.md) performs the GitHub sign-in and API key setup in one non-interactive command.
+
```
Ruari-Phipps: Probably remove this

src/poly/handlers/posthog.py
Ruari-Phipps, line 1: Do we need this update? I can't see any of these being used at the moment? Or have I missed something

src/poly/cli_commands/apikey.py
```diff
@@ -0,0 +1,512 @@
...
+    def say(self, fn: Callable[[str], None], message: str) -> None:
+        """Print via `fn`, unless --json is active - its output must be the only thing on stdout."""
+        if not self.output_json:
+            fn(message)
+
+    def emit(self, event: str, **extra: object) -> None:
+        """Capture a telemetry event tagged with the current distinct id."""
+        from poly.handlers.posthog import capture_event
+
+        capture_event(self.region, event, {**self.base_properties, **extra}, self.distinct_id)
```
Ruari-Phipps: Oh missed this. I thought we'd want this on the Auth0 side not ADK side

src/poly/cli_commands/apikey.py
```diff
@@ -0,0 +1,512 @@
...
+            only) more than one account was found and `account_id` was not given.
+    """
+    if account_id is not None:
+        return account_id
+    for _ in range(ACCOUNT_POLL_ATTEMPTS):
+        accounts = api.get_accounts_internal(
+            region=region, jwt_token=jwt_token, source=APIKEY_SOURCE
+        )
+        if accounts:
+            if region != "studio" and len(accounts) > 1:
```
Ruari-Phipps:
```suggestion
            if len(accounts) > 1:
```

src/poly/cli_commands/apikey.py
```diff
@@ -0,0 +1,512 @@
...
+        apikey_parser = subparsers.add_parser(
+            "apikey",
+            parents=[parents.verbose, parents.debug, parents.json],
+            help="One-shot setup for AI coding assistants.",
+            description=(
+                "One-shot setup: sign-in, account API key, POLY_API_KEY in your"
+                " environment.\n\n"
+                "Never prompts. Designed to be run by an AI coding assistant; fine to run"
+                " yourself.\n\n"
+                "To sign in interactively instead, use `poly login`.\n\n"
```
Ruari-Phipps: Again, this is a more limited version of `poly login` So not sure we should be positioning it like this

src/poly/cli_commands/apikey.py
```diff
@@ -0,0 +1,512 @@
...
+    return DEFAULT_KEY_NAME if region == "studio" else f"{DEFAULT_KEY_NAME}-{region}"
+
+
+class _Reporter:
+    """Quiet-when-json printing and telemetry, tied to the flow's distinct id.
+
+    A small stateful object rather than closures, since the distinct id
+    changes partway through the flow (from an anonymous id to the resolved
+    account id) and every step after that needs to see the new value.
+    """
```
Ruari-Phipps: Looks overcomplicated for just this flow

src/poly/cli_commands/apikey.py
```diff
@@ -0,0 +1,512 @@
...
+            account_id=args.account_id,
+            force=args.force,
+            output_json=args.json,
+            verbose=args.verbose,
+        )
+
+    @classmethod
+    def apikey(
+        cls,
+        region: str = "studio",
```
Ruari-Phipps: Think we shouldn't force this. Get them to specify and error if not present

src/poly/handlers/posthog.py
andymohajeri: Agreed - removed. Sign-up tracking moves to the Auth0 post-login Action; posthog.py is back to main.

src/poly/cli_commands/apikey.py
andymohajeri: Yes - moved to the Auth0 side. All ADK telemetry dropped.

src/poly/cli_commands/apikey.py
andymohajeri: Replaced the class with two small functions, _say and _fail.

src/poly/cli_commands/apikey.py
andymohajeri: Applied as suggested - no studio special case.

src/poly/cli_commands/apikey.py
andymohajeri: --region is now required with no default.

src/poly/cli_commands/apikey.py
andymohajeri: Reworded: it's for getting an API key for the PolyAI APIs / Dialog RSN, with setup/login as the ADK path.

docs/docs/reference/cli/apikey.md
andymohajeri: Reworded the same way; the page now leads with the API-key use case and points ADK setup at setup/login.

docs/docs/get-started/get-started.md
andymohajeri: Reframed the subsection around API keys for the APIs / Dialog RSN and pointed ADK setup at setup/login.

docs/docs/tooling/tooling.md
andymohajeri: Removed.

src/poly/handlers/posthog.py
andymohajeri: Bringing this back. Moved tracking off the Auth0 side: infra flagged an architectural preference for keeping Auth0 free of custom code, since it gets complicated to manage there. capture_event() here is new (this module previously only evaluated feature flags), used by apikey.py to fire one signup event.

src/poly/cli_commands/apikey.py
andymohajeri: Reversing this one. Infra stated an architectural preference for keeping Auth0 free of custom code since it gets complicated, so moved sign-up tracking back into apikey itself. Fires one apikey_signup event, gated to --region studio, off the is_first_login signal /jupiter/v1/authorise already returns.

jamesosullivan-polyai:
## Review

Ran `ruff check`, `ruff format --check`, and `pytest` — all clean (1992 tests passed). Tests were added for the new command, device flow, and env profile writer.

### Blocking

**Activation check validates the wrong key** ([`apikey.py`](https://github.com/polyai/adk/blob/andym/DEVP-690/implement-adk-command-for-sign-up/src/poly/cli_commands/apikey.py#L141-L170))

When a credential for the region already exists, `_save_credentials` deliberately keeps it and doesn't write the newly created key to disk. But `_wait_for_key_active` calls `api.get_accounts(region)`, which resolves its credential via `retrieve_api_key(region)` — i.e. whatever's already on disk, not the key just created. In this "existing credential kept" path (which the PR explicitly handles, for anyone who's already run `login`/`setup`), the poll silently checks the old, already-active key and reports `key_active: true` regardless of whether the actual key being exported is active yet. That defeats the documented purpose of the wait ("so a command run immediately afterward doesn't fail against a key the platform hasn't finished activating") and makes the `--json` `key_active` field unreliable in this common scenario.

`test_existing_studio_credential_is_not_overwritten` doesn't catch this — it only asserts `save_cred` wasn't called, not what the activation check validates.

Suggested fix: probe activation with the new key itself (e.g. call the account-keys list endpoint with the fresh key/JWT) rather than `get_accounts`, which depends on the on-disk credential.

### Minor

- `--json` error output always includes a full Python traceback, even without `--verbose`, unlike the human-readable path. No secrets leak, but worth gating for consistency or documenting as intentional.
- The multi-account check in `_resolve_account_id` fails fast on the first poll rather than retrying if a transient duplicate account briefly appears right after signup. Low risk, worth confirming with the API team whether that can happen.

### Nit

- Hardcoded Auth0 device client ID in `auth0_handler.py` — consistent with existing patterns in that file, not a secret, just flagging since "no hardcoded credentials" is on the checklist.

Everything else (style, docstrings, error handling, tests, security posture around key masking) looks solid — nice cleanup extracting the shared device-flow polling loop too.

andymohajeri:
Thanks for the review, James.

**Blocking - fixed in 78118c4.** You're right: `_wait_for_key_active` was calling `get_accounts(region)`, which authenticates via the on-disk credential, not the key this run actually created or reused. In the "existing credential kept" path that silently validated a different, already-active key. Added `get_accounts_with_key` (explicit-key auth, reusing `make_request`'s existing `headers` override, no new backend endpoint needed) and threaded the run's real `api_key` through the check instead. Added a regression test (`test_probes_the_actual_key_not_the_on_disk_credential`) and updated the existing `KeyActivationWait` tests to mock the new method.

One extra data point in case it's useful: checked the backend - the `active` column defaults to `true` immediately at key creation with no async flip, so this wait was likely never really about a DB-level activation delay, more about something downstream (auth-gateway/cache propagation) catching up. Probing with the real key over the real auth path seemed like the more correct fix either way, not just a patch for the specific bug.

**Traceback nit - also fixed in 78118c4.** Turned out to be simpler than "gate it" - since `_fail` already re-raises immediately under `--verbose`, the JSON branch is only ever reached when `verbose=False`, so the traceback was included unconditionally in exactly the one case it shouldn't be. Just dropped the field entirely; updated the contract test accordingly.

**Multi-account transient-duplicate retry and the hardcoded Auth0 client ID** - agreed these aren't code changes for this PR. Will raise the transient-duplicate question with the API team separately.

All tests pass (1995 total), ruff clean.

jamesosullivan-polyai (approved):

PR #326, opened 2026-09-23
perf: use the libyaml parser for YAML loads and add a large-project benchmark

Ruari-Phipps:
## Summary

YAML loads now use ruamel's libyaml-backed parser, and a new offline benchmark measures pull, status and push on a large multi-resource project. In the benchmark, total sync time falls by about 38% and files on disk come out byte-for-byte the same.

## Motivation

A large production project, with a 1.4 MB `variant_attributes.yaml` (about 1,200 variants) and a 400 KB `pronunciations.yaml` (about 3,300 rules), spends tens of seconds per pull or push parsing and dumping YAML in pure Python. On a CPU-limited service that was long enough to fail health checks.

Since ruamel.yaml 0.19, its C extension is an optional extra, so we have been running the pure-Python parser. This is the first of three perf changes. It also adds the benchmark the next two use to prove their output is unchanged.

## Changes

- Add `ruamel.yaml.clib` as a runtime dependency. `load_yaml` (`YAML(typ="safe")`) picks up the C parser automatically and keeps YAML 1.2 value resolution, so `yes`, `on`, `12:30` and `0777` read exactly as before. `dump_yaml` is unchanged: the round-trip emitter has no C implementation.
- The environment marker installs the extension only where prebuilt cp314 wheels exist: CPython on macOS, Linux (glibc and musl) and Windows x64/x86. It is skipped on Windows on ARM and on Python 3.15 or later until wheels ship. Where it is missing, ruamel falls back to pure Python, so installs never need a compiler. Free-threaded 3.14t builds can't be excluded with a marker and would build from source.
- Add `scripts/bench_large_project.py`. It builds a synthetic projection at production scale and runs init, force pull, pull with a remote edit, status after a local edit, and a dry-run push. Everything runs offline through the real code paths. `--digest-out` records a sha256 of every file after each phase, and `--profile PHASE` prints a cProfile report.
- Add `LoadYamlParserTests`. It checks that every fixture YAML file and an edge-case document (implicit types, dates, ` `, NBSP, emoji, anchors, CRLF, BOM) load identically with the C and pure-Python parsers. It also checks that the C parser is actually in use on platforms where the marker installs it.
- Regenerate `uv.lock` and `licenses.json`. Both gain one new MIT entry. The lock also catches up with the 0.60.0 version bump.

## Test strategy

I compared benchmark digests with and without `ruamel.yaml.clib` installed, and every file after every phase is identical. I also built the wheel, installed it into a fresh venv, and confirmed it resolves the extension and uses `CParser`.

## Checklist

- [x] `pytest` passes

## Screenshots / Logs

`uv run python scripts/bench_large_project.py`: 1,238 variants × 21 attributes (1.42 MB), 3,317 pronunciations (0.40 MB). Best of two runs on an M-series laptop:

| Phase | Pure Python | libyaml parser |
|---|---|---|
| init | 2.44 s | 2.08 s |
| force pull | 5.63 s | 3.05 s |
| pull (one remote edit) | 12.91 s | 7.79 s |
| status (one local edit) | 2.62 s | 1.17 s |
| dry-run push | 1.27 s | 1.31 s |
| **total** | **24.87 s** | **15.45 s** |

Most of the remaining pull time is per-rule dump→load round trips in `Pronunciation.save`, which the next PR removes.


src/poly/tests/utils_test.py
```diff
@@ -1309,5 +1315,85 @@ def test_write_json_file_preserves_unicode(self):
...
+  folded
+  text
+flow: {a: [1, 2], b: null}
+anchor: &shared
+  key: value
+alias: *shared
+"""
+
+
+def _libyaml_expected() -> bool:
```
andymohajeri: `_libyaml_expected()` reimplements the exact platform and version check from the `ruamel.yaml.clib` marker in `pyproject.toml`, by hand, in a second place. They agree today, but nothing keeps them in sync. If the marker ever changes, like excluding another platform, the test's copy could quietly drift and pass even though it's checking the wrong thing. Can this check live in one place that both sides use?

pyproject.toml
```diff
@@ -193,6 +193,9 @@ dependencies = [
     "protobuf>=4.21.0",
     "rich>=13.0.0",
     "ruamel.yaml>=0.18.0",
+    # libyaml-backed parser for YAML loads. Pinned to a version with prebuilt cp314 wheels;
+    # skipped where none exist (Windows on ARM, future Pythons) so installs never need a compiler.
+    "ruamel.yaml.clib==0.2.15; platform_python_implementation == 'CPython' and python_version < '3.15' and (sys_platform != 'win32' or platform_machine != 'ARM64')",
```
andymohajeri: This marker cuts off at `python_version < '3.15'`, but `requires-python` has no upper bound. The day this repo runs on 3.15, the C parser install gets silently skipped and everything falls back to pure Python, with no CI signal, since every clib-dependent test is conditionally skipped rather than required. Worth a tracking issue or reminder to re-pin once 3.15 wheels for `ruamel.yaml.clib` exist, so this whole perf series doesn't quietly regress on the next interpreter upgrade.

src/poly/tests/utils_test.py
Ruari-Phipps: Good catch. `_libyaml_expected()` now evaluates the `ruamel.yaml.clib` marker from the installed package metadata (`importlib.metadata.requires` + `packaging`), so `pyproject.toml` is the only place the check lives. I also changed the marker from excluding Windows ARM64 to listing the architectures 0.2.15 has wheels for, and the test didn't need to change.

pyproject.toml
Ruari-Phipps: Agreed. Rather than a reminder, I added `test_clib_marker_covers_running_python`, which evaluates the same marker for a platform we know has wheels. It only fails when the running Python falls outside the version bound, so bumping CI to 3.15 will fail with a message saying to re-pin `ruamel.yaml.clib`. I kept the upper bound: without it, pip on 3.15 would try to build 0.2.15 from source and need a compiler.

andymohajeri (approved):
Everything looks good

PR #327, opened 2026-09-23
fix: give each thread its own multi-resource file cache

Ruari-Phipps:
## Summary

`MultiResourceYamlResource._file_cache` becomes a thread-local dict instead of one dict shared by the whole process. Every existing call site stays the same, so single-threaded use (CLI, tests) behaves exactly as before.

## Motivation

A service that runs the ADK in-process crashlooped because its handlers ran the ADK inline on the event loop. The fix there moves ADK calls onto a worker thread, but it has to hold them to **one at a time**, because the shared cache isn't safe when two calls run together:

- `project.py` calls `_file_cache.clear()` at 7 points during pull and load. Another thread's writes that were only cached (`save_to_cache=True`) and not yet flushed are dropped.
- The merge path builds `original_file_contents` from every cache entry (`project.py:825-828`), so it picks up another thread's files.
- `write_cache_to_file()` flushes every entry, including another thread's files, into their temp dirs.

With one cache per thread, callers can run ADK calls in parallel threads safely. It's also a prerequisite for free-threaded Python.

## Changes

- `resource.py`: `_file_cache` is now a small descriptor, `_PerThreadFileCache`, that returns the calling thread's dict. The dict lives in a `threading.local`.
  - All uses, including `cls._file_cache.get/clear/setdefault`, iteration, subclasses and tests, keep working unchanged.
  - Subclasses still share the base class's cache within a thread. That matters because projects clear the cache via `MultiResourceYamlResource` while subclasses write via `cls`.
- Why `threading.local` and not a `ContextVar`: copied contexts share the same dict object, and on free-threaded 3.14 builds threads inherit their parent's context by default. Either would put the shared cache back.
- `resources_test.py`: new `MultiResourceFileCacheTests`. They check that another thread starts with an empty cache, that clearing or flushing in another thread leaves this thread's entries and files alone, and that subclasses share the calling thread's cache.

## Test strategy

- [x] Added/updated unit tests. 3 of the 4 new tests fail on `main` and pass with this change. The fourth, which checks that subclasses share the cache, passes on both.

## Checklist

- [x] `pytest` passes (2052 passed)


andymohajeri (approved):
looks good

PR #329, opened 2026-09-23
perf: index multi-resource lookups and reuse mapping tables

Ruari-Phipps:
## Summary

Removes three sources of repeated full-list work on large projects: finding an entry in a multi-resource YAML file by name, rebuilding lookup tables from resource mappings for every resource read, and saving the whole project config on every `api_handler` access.

## Motivation

On a large production project (about 1,238 variants with 21 attributes each, plus 3,317 pronunciation rules), `pull`, `status` and `push` spend a lot of time on work that grows quadratically with project size:

- Every lookup by name in a multi-resource file (`_find_matching`) scanned the whole list and called `clean_name` on every entry. Each command does one lookup per resource, so the cost is O(N²).
- Every pronunciation read copied the whole rule list and scanned it to find the entry at a known index.
- Every resource read rebuilt its reference-lookup dict, and every variant attribute rebuilt its variant id/name map, from the full mappings list (thousands of entries). The mappings list is the same for the whole command.
- The `api_handler` property called `save_config()` on every access, because `branch_id` is always set. Each call serialises the entire project, and a push reads the property several times.

The benchmark script comes from #326. This PR depends on #326 only for that script.

## Changes

- `MultiResourceYamlResource` keeps an index per `(top_level_name, resource_key)` holding the list it was built from, each entry's clean name, and the position of the first entry for each clean name. The index is only used while it refers to the same list object at the same length, so when `_file_cache` reloads or clears a file (which creates new lists) the index is rebuilt. `save` and `delete_resource` keep the index up to date, and first-match semantics are unchanged when two entries share a clean name.
- New `_entry_key` classmethod for the clean name an entry is matched by. `KeyphraseBoosting` overrides it to read resolved YAML scalars as text, as its discovery already does. A hand-typed `keyphrase: 2024` or `keyphrase: true` can now be looked up. Before this change, `clean_name` raised `TypeError` on that entry, so every lookup that reached it failed.
- `Pronunciation.read_from_file` and `read_local_resource` read the rule directly at the path's index, with a bounds check. A missing index still raises `FileNotFoundError`.
- New `resource_utils.memo_for_mappings(mappings, key, build)`: a single-slot cache of tables derived from one mappings list. A cached table is reused while calls pass the same list object at the same length. An empty or missing list always calls `build()`. It is used by the reference replacer and by the variant id/name tables in `Variant.validate`, `VariantAttribute.validate`, `to_pretty_dict` and `from_pretty_dict`.
- `sync_ids_with_parent` builds its combined mappings list once, before the loop, instead of on every iteration.
- `api_handler` saves the config only when the handler's branch differs from `project.branch_id`. Every branch change is still persisted:
  - A branch set inside the handler (initial branch resolution, fallback to `main` when the branch was deleted, auto-created branch on push) is saved on the next property access.
  - Branch switches and branch creation driven by the project are followed by an explicit `save_config()`.
  - One timing difference: after `poly branch switch`, the new branch is saved by the pull's explicit `save_config()` rather than by a property access. If that pull fails part-way, the status file keeps the previous branch; before, it recorded the new branch next to the old resources.
  - Saves that happened only as a side effect of reading the property are gone. They also wrote migration flags and `last_updated` early; both are still written by the next explicit save, and migrations are designed to re-run while their flag is absent.

## Test strategy

New tests read and write real files in a temp directory:

- Index: duplicate clean names (first match wins for read, save and delete); delete then append; a save that changes an entry's name; numeric and boolean keyphrases; reload after the file changes on disk; variants and attributes interleaved in `variant_attributes.yaml`.
- Pronunciation reads by index, including out-of-range indexes.
- Memo: reuse, separate tables per key, invalidation when the list is extended, a rebuild for a different list of the same length, empty lists that are never cached, and reference swaps equal to the unmemoised builder across flow scopes and both directions. Variant validation and pretty-dict round trips pick up a variant appended after an earlier call.
- `api_handler`: no save while the branch is unchanged, and exactly one save when the handler's branch changes.

The benchmark writes identical file digests on `main` and on this branch after every phase.

## Checklist

- [x] `pytest` passes

## Screenshots / Logs

`scripts/bench_large_project.py` from #326 at its default scale (1,238 variants × 21 attributes, 3,317 pronunciations). Both columns use the pure-Python YAML parser because libyaml is not on `main`. Each figure is the best of five runs, and `main` and branch runs were interleaved.

| Phase | `main` | This PR |
|---|---|---|
| init | 2.47 s | 2.28 s |
| force-pull | 6.78 s | 5.11 s |
| pull | 14.87 s | 11.59 s |
| status | 2.52 s | 1.72 s |
| push | 1.30 s | 0.50 s |
| total (best single run) | 29.15 s | 22.32 s |

Profile of `status` (cProfile, cumulative):

| Function | `main` | This PR |
|---|---|---|
| `MultiResourceYamlResource._find_matching` | 0.44 s | 0.004 s |
| `Pronunciation.read_local_resource` (own time) | 1.07 s | 0.01 s |
| `_build_reference_replacer` | 0.08 s | 0.003 s |

In the benchmark's `push`, `save_config` goes from 3 calls (0.57 s under the profiler) to 0.

The remaining `pull` time is mostly YAML parsing and per-rule pronunciation saves, which are handled separately.


src/poly/resources/resource.py
```diff
@@ -502,13 +502,26 @@ def _parse_multi_resource_path(file_path: str) -> tuple[str, list[str]]:
...
 class MultiResourceYamlResource(YamlResource, ABC):
     """Abstract base class for a resource that is stored in a single YAML file with multiple resources."""
 
     # Class-level cache: true_file_path -> (mtime, top_level_yaml_dict). Invalidated on write; refreshed when mtime differs.
     _file_cache: ClassVar[dict[str, tuple[float, dict]]] = {}
 
+    # (top_level_name, resource_key) -> (yaml_list, keys, positions): the clean name of each entry
+    # and the index of the first entry per clean name. Only valid for the exact list object it was
+    # built from, so a reload or cache clear (which yields new lists) drops it.
+    _match_index: ClassVar[dict[tuple[str, str], tuple[list, list[str], dict[str, int]]]] = {}
```
andymohajeri: `_file_cache` gives each thread its own copy of the dict of loaded YAMLs (from #327). Looks like the two new caches added in #329 for the speedup, `_match_index` and `memo_for_mappings`, didn't get that same treatment. They're still one shared, process-wide cache each. Can we apply #327's fix here too, so each thread gets its own private copy?

andymohajeri: `_match_index`'s staleness check only catches a new list object or a change in length. It won't catch an existing entry's key field being edited in place by any code path other than `save()` or `delete_resource()`. I checked every `MultiResourceYamlResource` subclass in the repo and nothing does that today, so this isn't live, but nothing stops a future migration script from doing it either. Can we add a comment on the `_match_index` declaration spelling out that only `save()` and `delete_resource()` are allowed to mutate a cached entry in place?

src/poly/project.py
```diff
@@ -143,9 +143,11 @@ def api_handler(self) -> AgentStudioInterface:
                 self.project_id,
                 self.branch_id,
             )
-        self.branch_id = self._api_handler.branch_id
-        if self.branch_id:
-            self.save_config()  # To save branch if it changed (e.g. deleted in remote)
+        handler_branch_id = self._api_handler.branch_id
+        if handler_branch_id != self.branch_id:
```
andymohajeri: This moves branch persistence from "save on every property read" to "save only when the branch actually changed." You call out yourself that if a pull fails partway, the status file now keeps the old branch next to the old resources instead of recording the new branch too early. That sounds like the safer outcome. `ApiHandlerPropertyTest` covers the save-on-change behavior in isolation, but there's no test for the actual failed-pull-partway scenario. Can we get a sign-off from whoever owns branch-switch recovery, and maybe a test that fails a pull midway and checks the status file still matches what's on disk?

PR #330, opened 2026-09-23
feat: carry deployment mode over when duplicating a project

paul2048:
## Summary

`poly project duplicate` now gives the copy the same deployment mode (Simple, Branches or Sub-branches) as the project it was duplicated from.

## Motivation

Projects duplicated with the ADK always came out in Simple mode, whatever mode the source used, and the only way to change it was in Agent Studio. New projects start in Simple by design, so the ADK now copies the source's mode onto the duplicate itself.

## Changes

- After duplicating, `poly project duplicate` reads `config.deployment_mode` from both projects. If they differ, it updates the copy to match the source.
- If that step fails, the duplicate still succeeds: the console shows a warning, and `--json` output gets a `deployment_mode_error` key.
- Adds `PlatformAPIHandler.update_project` (`PATCH /adk/v1/accounts/{account_id}/projects/{project_id}`) and `AgentStudioInterface.set_deployment_mode`.

## Test strategy


Ruari-Phipps (approved):
Nice! Lgtm!

